PR #90, opened 2026-04-12
feat: add repo-level default override setup

### LINKED ISSUE #89: Repo-level default override setup
## Summary

`compozy setup` already installs skills and global reusable agents, but it could also offer an optional repo-level step to override execution defaults and save them to `.compozy/config.toml`.

## Why

Compozy already has built-in defaults, so users do not need a config file for normal execution. The value here is convenience: setup can give users a guided way to change those defaults and save them to the repo without editing TOML by hand.

The README already documents workspace config and gives a concrete `[defaults]` example:

```toml
[defaults]
ide = "codex"
model = "gpt-5.4"
reasoning_effort = "medium"
```

That makes setup the natural place to ask whether the user wants to override the built-in defaults for this repo and save those choices automatically.

Right now the flow is:

- Compozy uses built-in defaults automatically
- the README explains how to override them in `.compozy/config.toml`
- users who want repo-specific defaults still have to create or edit that file manually

An optional setup step would reduce friction for that manual step.

## Proposed Behavior

During interactive `compozy setup`:

- after agent/skill selection, ask whether to override repo-level runtime defaults
- default that answer to `No`
- if the user opts in, write the chosen overrides to `.compozy/config.toml` in the current workspace only
- do not prompt for this during `compozy setup --global`
- do not write config if setup installation fails

Suggested fields for the first version:

- `defaults.ide`
- `defaults.model`
- `defaults.reasoning_effort`

## Recommendation Details

- `ide` should be selected from the supported runtime list
- `model` should be recommended from the existing per-IDE runtime default mapping
- `reasoning_effort` should use the same built-in default already used by runtime config

## Expected Outcome

Users can finish setup with either:

- no repo config changes, and Compozy keeps using its built-in defaults
- or repo-specific overrides saved to `.compozy/config.toml` without editing the file manually

That keeps setup aligned with the current README and preserves the existing workspace-only config model.

## Notes

- This should stay workspace-scoped. Global setup should not mutate repo config.
- Explicit CLI flags should still override config as they do today.

camopy:
## Summary

Add an optional repo-level config step to `compozy setup` so interactive setup can override built-in runtime defaults and save those overrides to `.compozy/config.toml`.

## Why

Compozy already works without a repo config file because it has built-in defaults. This change makes repo-level overrides easier to set.

The README already documents workspace config and shows recommended `[defaults]` values, but setup previously stopped at skill installation. This change closes that gap by letting users change and save repo-specific defaults from the setup flow instead of editing TOML manually.

## What Changed

- added an interactive repo-level defaults step to `compozy setup`
- defaulted the override prompt to `No`
- limited the config write to workspace scope only
- skipped the prompt entirely for `compozy setup --global`
- only persist defaults after setup completes with zero failures
- restored IDE selection as a dropdown of supported runtimes
- used the runtime registry's per-IDE default model as the recommendation source
- reused the shared built-in default reasoning effort instead of a setup-local hardcoded value
- preserved existing config sections when writing the selected defaults

## Config Written

When the user opts in, setup updates:

```toml
[defaults]
ide = "..."
model = "..."
reasoning_effort = "..."
```

## Behavior Notes

- repo config is written to `.compozy/config.toml`
- users who opt out keep the normal built-in defaults with no config change
- global setup does not prompt for repo defaults and does not modify workspace config
- if the current model matches the previous recommended model, changing IDE updates the model recommendation to the new IDE default
- if the repo already has a custom model, setup preserves it

## Testing

- added coverage for registry-backed IDE default model resolution
- added coverage for preserving unrelated config sections on write
- added coverage for skipping repo-default prompts in global mode
- added coverage for not persisting config when setup installation fails

## Verification

```text
make verify
```

Passed locally.

Closes #89 

## Summary by CodeRabbit

* **New Features**
  * Interactive runtime defaults during setup (choose preferred IDE, model, reasoning effort) with preview and persistence to workspace config.
  * Added default reasoning level ("medium") and a lookup for IDE-specific default models.
  * Config read/write updated to omit empty sections when saving and to explicitly handle load/write semantics.

* **Tests**
  * Added tests covering runtime-default selection, persistence, config read/write, prompting flows, and context/error propagation.

### DIFF
--- a/internal/cli/commands_test.go
+++ b/internal/cli/commands_test.go
@@ -89,6 +89,7 @@ func TestNewExecCommandRegistersExtensionsFlag(t *testing.T) {
 	flag := cmd.Flags().Lookup("extensions")
 	if flag == nil {
 		t.Fatal("expected exec command to register --extensions")
+		return
 	}
 	if flag.DefValue != "false" {
 		t.Fatalf("expected --extensions default false, got %q", flag.DefValue)
@@ -102,6 +103,7 @@ func TestNewTasksRunCommandDefaultsAttachModeToAuto(t *testing.T) {
 	flag := cmd.Flags().Lookup("attach")
 	if flag == nil {
 		t.Fatal("expected --attach flag")
+		return
 	}
 	if flag.DefValue != attachModeAuto {
 		t.Fatalf("expected --attach default %q, got %q", attachModeAuto, flag.DefValue)
@@ -133,6 +135,7 @@ func TestNewTasksCommandDoesNotRegisterRunMultiple(t *testing.T) {
 		flag := run.Flags().Lookup("attach")
 		if flag == nil {
 			t.Fatal("expected run --attach flag")
+			return
 		}
 		if flag.DefValue != attachModeAuto {
 			t.Fatalf("expected --attach default %q, got %q", attachModeAuto, flag.DefValue)
@@ -147,6 +150,7 @@ func TestNewTasksRunCommandRegistersRecursiveFlag(t *testing.T) {
 	flag := cmd.Flags().Lookup("recursive")
 	if flag == nil {
 		t.Fatal("expected --recursive flag on tasks run")
+		return
 	}
 	if flag.Shorthand != "r" {
 		t.Fatalf("expected --recursive shorthand %q, got %q", "r", flag.Shorthand)
@@ -166,6 +170,7 @@ func TestReviewsFixCommandDefaultsTUIToTrue(t *testing.T) {
 	flag := cmd.Flags().Lookup("tui")
 	if flag == nil {
 		t.Fatal("expected --tui flag")
+		return
 	}
 	if flag.DefValue != "true" {
 		t.Fatalf("expected --tui default true, got %q", flag.DefValue)
--- a/internal/cli/form_test.go
+++ b/internal/cli/form_test.go
@@ -404,6 +404,7 @@ func TestTaskRunRuntimeFormPreseedsConfiguredTypeRules(t *testing.T) {
 		}
 		if form == nil {
 			t.Fatal("expected task runtime form")
+			return
 		}
 		if !slices.Contains(form.selectedTypes, "demo::backend") {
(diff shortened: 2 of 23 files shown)


pedronauck (changes requested):
Thanks for the PR. I like the direction here: integrating repo-level runtime defaults into `compozy setup` closes a real usability gap, and I confirmed the branch passes `make verify` locally on commit `8341c23`.

I do not think this is ready to merge yet, because the new flow conflicts with existing workspace/setup semantics in a few important ways:

1. Project-scope setup and workspace config can target different directories.
`resolveRuntimeDefaultsPlan()` / `runtimeDefaultsPath()` in `internal/cli/setup.go` now resolve `.compozy/config.toml` through workspace discovery, but project-scope skill installation in `internal/setup/install.go` still resolves paths relative to the current working directory. If `compozy setup` is run from a nested directory, this PR can write config in the workspace root while installing skills under the nested directory's `.agents/skills`, creating split state across two locations.

2. The runtime-defaults prompt ignores the agents selected in the same setup run.
The install plan already knows `selectedAgents`, but the defaults step still falls back to Codex when there is no existing config. That means a user can install only Claude or Cursor, accept the suggested defaults, and still end up with `defaults.ide = "codex"`. Because workspace defaults are later applied directly to command state in `internal/cli/workspace_config.go`, and bundled-skill preflight validates the selected IDE in `internal/cli/skills_preflight.go`, the next `start` / `fix-reviews` invocation can fail with missing Codex skills immediately after setup.

3. The "optional" defaults step is not actually optional when config parsing fails.
Before the user answers the opt-in prompt, `runtimeDefaultsPath()` resolves the workspace, and `workspace.Resolve()` already loads/parses `.compozy/config.toml`. A malformed or unreadable config now aborts interactive `compozy setup` even if the user would choose `No` and skip repo defaults entirely. That looks like a regression in the current setup flow.

Recommendation: I would address these before merging. The feature itself is useful and not duplicative, but the current implementation introduces scope inconsistencies and post-setup failure modes.

pedronauck (changes requested):
@camopy there are some conflicts to solve here

PR #96, opened 2026-04-14
feat: add optional sound notifications on run lifecycle events

brunomguimaraes:
## Summary

Adds optional audio cues that play when an ACP run reaches a terminal state — a small ergonomic win for operators who step away from long-running sessions and want to know without checking the terminal whether the model finished cleanly or failed.

The feature is **opt-in** via a new `[sound]` section in `.compozy/config.toml` and ships **disabled by default**, so it costs nothing for users who don't want it.

```toml
[sound]
enabled = true
on_completed = "glass"   # preset name OR absolute path
on_failed = "basso"      # played on run.failed and run.cancelled
```

https://github.com/user-attachments/assets/cf313104-cd17-4c4b-a1f3-5188e48b2e7a

Presets resolve to platform-native files at play time (`afplay` on darwin, `paplay` on linux, `powershell` on windows). Absolute paths bypass preset lookup so users can ship custom sounds.

## Why

- A long ACP session can run for minutes; switching back to the terminal only to check a status string is friction.
- A 500ms beep at the boundary is enough to surface the result without requiring focus.
- Keeping it strictly opt-in means CI, headless servers, and silence-loving users see no behavior change.

## Design notes

**Synchronous playback at the emission point.** The first iteration used a goroutine subscribed to the run event bus, but cleanup raced the terminal event and could silently drop the very sound it was meant to play. The final design exposes `sound.Notify(ctx, cfg, kind, logger)` which plays on the caller's goroutine right after `emitRunTerminalEvent` (executor) and `submitRuntimeEvent` (exec). It blocks for ~500ms on the audio command, then returns. Single-flight, deterministic, no concurrency surface.

**Cross-platform isolation via build tags.** `player_unix.go` (`//go:build !windows`) selects `afplay`/`paplay` based on `runtime.GOOS`; `player_windows.go` shells out to `powershell` + `System.Media.SoundPlayer`. Other unix variants gracefully fall back to a `Noop` player.

**Test seam.** `os/exec` is hidden behind a `commandRunner` interface so unit tests never shell out — they assert the resolved command name and args without producing audio.

**TOML-only for v1.** No CLI flag was added; per-project sound feels like a project preference rather than a per-invocation toggle. Easy to add later if there's demand.

## Files touched

| Area | Files |
|---|---|
| New package | `internal/core/sound/{player,runner,presets,player_unix,player_windows,subscriber}.go` (+ tests) |
| Config schema | `internal/core/workspace/config_types.go`, `config_validate.go`, `config_test.go` |
| Runtime config | `internal/core/model/runtime_config.go`, `model_test.go` |
| Executor wiring | `internal/core/run/executor/execution.go` (synchronous `notifySoundForKind` after `emitRunTerminalEvent`) |
| Exec wiring | `internal/core/run/exec/exec.go` (`notifySoundForKind` inside `submitRuntimeEvent`) |
| Plumbing | `internal/core/run/internal/runshared/config.go`, `internal/core/api.go`, `internal/cli/state.go`, `internal/cli/workspace_config.go` |
| SDK alignment | `sdk/extension/types.go` (mirrors `model.RuntimeConfig` per the existing `compat_test` contract) |

## Test plan

- [ ] `make verify` passes (1848 tests, 0 lint, build green)
- [ ] New tests cover preset resolution per OS, `Notify` for every terminal kind, silent kinds, nil player, nil context, playback errors logged not panicked
- [ ] Extended `TestRuntimeConfigApplyDefaults` to cover the sound default branch (presets only fill when `SoundEnabled` is true; explicit values are preserved)
- [ ] New `TestLoadConfig` cases for `[sound]` parsing, default-disabled state, and empty-string rejection
- [ ] Manual: `compozy exec --dry-run --tui=false "hello"` plays Glass on darwin
- [ ] Manual: `compozy exec --tui --dry-run "hello"` plays Glass with the cockpit visible
- [ ] Manual: feature is silent when `enabled = false` or the section is absent
- [ ] Maintainer to verify on linux (`paplay` + freedesktop sounds) — only tested darwin

## Not included

- "Awaiting input" event kind. The current run lifecycle has no native "blocked on user feedback" event; adding one is a larger surface change and didn't seem worth bundling. Happy to follow up in a second PR if there's interest.
- Job-level beeps. The wiring is intentionally restricted to the three run-level events to avoid a 50-beep batch run.
- Volume / debounce controls. Easy to add if real-world usage shows a need.

## Summary by CodeRabbit

* **New Features**
  * Sound notifications for run completion/failure across Windows, macOS, and Linux.
  * New [sound] project config to enable/disable notifications and set completion/failure sounds.
  * Configurable defaults applied when enabled (defaults: glass for success, basso for failure); preset names are trimmed and preserved when explicitly set.
  * Preset list: glass, basso, ping, hero, funk, tink, submarine.
  * Runtime SDK exposes sound-enabled fields for extensions.

### DIFF
--- a/internal/cli/state.go
+++ b/internal/cli/state.go
@@ -43,6 +43,9 @@ type runtimeConfig struct {
 	explicitRuntime  model.ExplicitRuntimeFlags
 	includeCompleted bool
 	includeResolved  bool
+	soundEnabled     bool
+	soundOnCompleted string
+	soundOnFailed    string
 }
 
 type execConfig struct {
@@ -317,6 +320,10 @@ func (s *commandState) buildConfig() (core.Config, error) {
 		Timeout:                timeoutDuration,
 		MaxRetries:             s.maxRetries,
 		RetryBackoffMultiplier: s.retryBackoffMultiplier,
+
+		SoundEnabled:     s.soundEnabled,
+		SoundOnCompleted: s.soundOnCompleted,
+		SoundOnFailed:    s.soundOnFailed,
 	}, nil
 }
 
--- a/internal/cli/workspace_config.go
+++ b/internal/cli/workspace_config.go
@@ -97,6 +97,7 @@ func commandPath(cmd *cobra.Command) string {
 }
 
 func (s *commandState) applyProjectConfig(cmd *cobra.Command, cfg workspace.ProjectConfig) {
+	applySoundConfig(s, cfg.Sound)
 	applyConfig(cmd, "ide", cfg.Defaults.IDE, func(val string) { s.ide = val })
 	applyConfig(cmd, "model", cfg.Defaults.Model, func(val string) { s.model = val })
 	applyConfig(cmd, "format", cfg.Defaults.OutputFormat, func(val string) { s.outputFormat = val })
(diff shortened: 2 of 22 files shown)


pedronauck:
this is a nice feature, there's a reason it's still in draft mode?

brunomguimaraes:
> this is a nice feature, there's a reason it's still in draft mode?

Yes, I haven't had the opportuity to fully test in other OS

pedronauck:
if you tested on macOS is fine for this first version @brunomguimaraes

pedronauck:
@brunomguimaraes just fix the coderabbit review and we can merge it

brunomguimaraes:
> @brunomguimaraes just fix the coderabbit review and we can merge it

Done @pedronauck , FYI the pre-existing TestHostAPITasksCreateRoundTripsMethodAndParams race in upstream sdk/extension deserves a separate fix PR regardless of what CodeRabbit says about this, it's unrelated to the sound work. I went ahead and created #104

PR #128, opened 2026-04-24
feat: update Codex model to gpt-5.5

marcioaltoe:
## Summary

- Default Codex/Droid runtime references to `gpt-5.5` and update docs, skill references, release notes, examples, scripts, and tests accordingly.
- Add Codex ACP compatibility handling for `gpt-5.5`, including an upgrade hint for `@zed-industries/codex-acp >= 0.12.0`.
- Remove the fixed `gpt-5.4` fallback recommendation and use generic supported-model guidance instead.

## Stack

This PR is stacked on #127 (`ma/makefile-bun-min-version`) so the fix PR stays focused and this PR contains only the Codex model update commits.

## Validation

- `make verify`
- `rg -n "gpt-5\\.4|gtp-5\\.4" . --glob '!bun.lock' --glob '!go.sum' --glob '!go.mod'` returns no active-code/test/docs matches outside historical archived artifacts checked earlier.

## Summary by CodeRabbit

* **New Features**
  * Runtime model compatibility checks for the Codex runtime.
  * Optional live model availability test for Codex (runs when enabled).

* **Updates**
  * Default Codex model bumped to gpt-5.5.
  * CLI help/examples and workspace defaults updated to new model identifiers.
  * Runtime/task model mappings and alias resolutions refined, including faster-variant mappings.

* **Bug Fixes**
  * Improved error messages with clear upgrade/install guidance for incompatible runtimes.

PR #127, opened 2026-04-24
fix: harden runtime activity and version handling

marcioaltoe:
## Summary

- Treat `BUN_VERSION` as a minimum supported Bun version instead of requiring an exact pin.
- Replace repeated task-mode string literals with `runModeTask` to satisfy lint and keep daemon query code consistent.
- Keep ACP activity marked active while session updates are being submitted, with tests wrapped in the repository's `Should...` subtest style.
- Normalize `codex/<model>` aliases to bare Codex model ids before sending `SetSessionModel` to the native Codex ACP runtime.
- Compare update notifications against released base versions, so git-describe builds like `v0.1.12-15-g834fec6` do not prompt users to downgrade to `0.1.12`.
- Preserve prerelease identifiers such as `1.2.3-1-gamma` unless the git-describe suffix contains a plausible short SHA.
- Publish SDK extension initialized state before the initialize response is observable, fixing the CI race in `TestHostAPIRunsStartReturnsRunHandle`.

## Root Cause

- The Bun gate compared exact strings instead of semver-compatible minimums.
- The update notifier parsed git-describe versions as semver prereleases, making a build ahead of `v0.1.12` appear older than release `0.1.12`.
- Native Codex accepts bare model ids such as `gpt-5.5`; provider-qualified `codex/gpt-5.5` is rejected for ChatGPT-account Codex sessions.
- The SDK initialize handler wrote the initialize response before setting `initialized`, allowing the host to issue a Host API call that could be rejected as not initialized.

## Testing

- [x] `bun x playwright install --with-deps chromium` (local equivalent for CI `bunx` step)
- [x] `go test ./sdk/extension -race -run 'TestHostAPIRunsStartReturnsRunHandle|TestHandleInitializePublishesInitializedStateBeforeResponse' -count=1000`
- [x] `go test ./internal/core/agent ./internal/update ./internal/core/run/internal/acpshared -race`
- [x] `CGO_ENABLED=1 make verify`

## Summary by CodeRabbit

* **Chores**
  * Bun version check now enforces a minimum required version with clearer "or newer / at least" messaging.
  * Version parsing now strips git-describe suffixes before semver comparison.
  * Runtime model strings normalized to trim whitespace and remove provider prefixes where applicable.
* **Reliability**
  * Activity tracking counts nested/concurrent work; handlers mark activity for the full lifecycle to avoid premature idle detection.
* **Bug Fixes**
  * Task/document kind handling normalized across components.
  * Extension initialization ordering updated to set internal state before responding.
[...]

PR #129, opened 2026-04-27
fix: infer task type during migrate

marcioaltoe:
## Summary
- infer a valid v2 task type during migrate for legacy `feature` / `feature implementation` tasks
- use legacy `domain` as a constrained fallback only when the direct type remap is ambiguous
- keep explicit unmapped follow-up only when inference is still unsafe or disallowed by the workspace registry

## Why
`compozy migrate` could emit `type: ""`, which then caused `compozy sync` to fail on the migrated workflow. This fixes the migration path at the source instead of relying on post-migration manual edits.

## Notes
- stacked on top of #128 to minimize conflicts with the currently open PR chain
- this PR should merge after #128, which already targets #127

## Verification
- `go test ./internal/core/migration ./internal/cli -run 'TestMigrate|TestTasksValidateCommandExecuteDirectCoversFailureAndSuccess|TestMigrateCommandExecuteDirectReportsUnmappedTypeFollowUp|TestMigrateCommandPrintsUnmappedTypeSummaryAndValidateFailsUntilFixed'`
- `make verify`

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation and parse failures are now reported as 422 Unprocessable Entity with clearer validation messages.
  * Error wrappers now preserve original error identity, improving diagnostics and client-facing error consistency.

* **New Features**
  * Migration preserves legacy domain context and auto-inferrs task types from legacy domain metadata.
  * Migrate output no longer prompts for a manual fix for unmapped types.


marcioaltoe:
@coderabbitai review

marcioaltoe:
@coderabbitai review

PR #153, opened 2026-05-14
feat: discover task files recursively in nested subdirectories

pblobp:
## Summary

- Adds opt-in `--recursive` / `-r` flag (default `false`) to `compozy tasks run` so workflows can keep `task_NNN.md` under nested feature subdirectories instead of flattening every task into the slug root. Flat (non-`--recursive`) runs are byte-identical to today's behavior on every code path I traced.
- Walker swaps `os.ReadDir` for `filepath.WalkDir` with a hardcoded skip predicate (`.`-prefix, `_`-prefix, `reviews-*`, `adrs/`, `memory/`). Results are emitted as forward-slash relative paths, sorted root-first then per-subdir alphabetical, numeric within each group.
- `tasks/store` accepts validated forward-slash relpaths via `*os.Root` subpath resolution; `_meta.md` totals (`countTasks`, `SnapshotTaskMeta`) and `CompleteNonTerminalTasks` are always recursion-aware so a flat run never under-reports stray nested files.
- Workflow memory is scoped per-subdir: tasks at `features/auth/task_01.md` write to `memory/features/auth/MEMORY.md`. Reads use closest-ancestor inheritance; writes always target the task's immediate scope so a sub-feature note never silently bleeds into an ancestor. Keeps any one `MEMORY.md` well below the 12 KB / 150-line soft cap on deep recursive runs.
- Surfaces wired across `--recursive` flag, runtime-override DTO, `[tasks.run] recursive` TOML key, and interactive task-runtime form.

### Design decisions

- **Opt-in flag + sibling Go function**: a new `ReadTaskEntriesRecursive` lives next to `ReadTaskEntries` instead of widening the existing signature. Avoids forcing a coordinated extension SDK release.
- **Directory-grouped numeric ordering**: root tasks first, then each subdirectory alphabetically, numerically within each group. Skip predicate is hardcoded (dot-prefix, underscore-prefix, `reviews-*`, `adrs/`, `memory/`).
- **Recursion-agnostic meta + bulk completion**: `countTasks` / `SnapshotTaskMeta` / `CompleteNonTerminalTasks` always recurse so `_meta.md` totals stay reliable across flat and recursive runs. `MarkTaskCompleted` accepts validated relative paths via `os.OpenRoot` for defense-in-depth.

### Always-recursive behavior to call out in release notes

Two surfaces intentionally ignore the flag and always scan recursively:

- `tasks.countTasks` / `tasks.SnapshotTaskMeta` / `tasks.CompleteNonTerminalTasks`
- `tasks.Validate` (preflight)

For users who already have stray nested `task_*.md` files under a slug but don't pass `--recursive` to runs, `_meta.md` totals and preflight validation will now surface those files. Users with no nested files see zero behavior change.

## Test plan

- [x] `make verify` (fmt + lint + test + build): 0 issues, 3140 tests pass, build OK
- [x] `internal/core/memory/store_test.go` — 8 new tests for per-subdir bootstrap, root fallback, ancestor isolation on writes, closest-ancestor on reads, append-bootstraps-template
- [x] `internal/core/tasks/walker_test.go` — recursive discovery + skip predicate + nested-memory regression
- [x] `internal/core/tasks/store_test.go` — `MarkTaskCompleted` accepts relpaths; rejects path escape
- [x] `internal/core/plan/prepare_test.go` — nested task flows through planner with scoped workflow memory
- [x] `internal/core/extension/host_reads_test.go` / `host_writes_test.go` — extension SDK semantics widened additively (bare `MEMORY.md` still targets root)
- [x] `internal/cli/root_test.go::TestRunPreparedStartFlatValidationIgnoresNestedDrafts` — flat-mode regression guard
- [x] `internal/core/workspace/config_test.go` — TOML `[tasks.run] recursive = true` round-trips

## Summary by CodeRabbit

* **New Features**
  * Added a --recursive (-r) option for task discovery so "tasks run" can include task files in subdirectories; configurable via CLI or workspace config.
  * CLI help updated to document recursive discovery and how it interacts with other sync/scopes.

* **Behavior Changes**
  * Preflight validation and task selection now respect the recursive setting, affecting which nested tasks are discovered and validated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/compozy/compozy/pull/153)

### DIFF
--- a/.agents/skills/compozy/SKILL.md
+++ b/.agents/skills/compozy/SKILL.md
@@ -73,7 +73,7 @@ For a detailed step-by-step walkthrough of each phase, read `references/workflow
 | **Workflow Execution** | | |
 | `compozy daemon` | Manage the home-scoped daemon lifecycle | `start`, `status`, `stop` |
 | `compozy workspaces` | Inspect and manage daemon workspace registrations | `list`, `show`, `register`, `unregister`, `resolve` |
-| `compozy tasks run` | Execute PRD task files through the daemon | `--name`, `--attach`, `--ui`, `--stream`, `--detach`, `--task-runtime` |
+| `compozy tasks run` | Execute PRD task files through the daemon | `--name`, `--recursive` / `-r`, `--attach`, `--ui`, `--stream`, `--detach`, `--task-runtime` |
 | `compozy exec` | Execute an ad hoc prompt | `--agent`, `--format`, `--prompt-file`, `--tui`, `--persist`, `--run-id` |
 | `compozy runs` | Attach, watch, and purge daemon-managed runs | `attach`, `watch`, `purge` |
 | **Review** | | |
@@ -172,6 +172,7 @@ types = ["frontend", "backend", "docs", "test", "infra", "refactor", "chore", "b
 
 [tasks.run]
 include_completed = false
+recursive = false
 
 [fix_reviews]
 concurrent = 2
--- a/.agents/skills/compozy/references/cli-reference.md
+++ b/.agents/skills/compozy/references/cli-reference.md
@@ -57,6 +57,7 @@ Execute PRD task files sequentially from a workflow directory through the shared
 | --- | --- | --- | --- |
 | `--name` | string | | Task workflow name (resolves to `.compozy/tasks/<name>`) |
 | `--include-completed` | bool | false | Include tasks already marked as completed |
+| `--recursive`, `-r` | bool | false | Discover `task_NNN.md` files in nested subdirectories of the workflow root |
 | `--skip-validation` | bool | false | Skip task metadata preflight check |
 | `--force` | bool | false | Continue after validation fails in non-interactive mode |
 | `--attach` | string | auto | Attach mode: auto, ui, stream, detach |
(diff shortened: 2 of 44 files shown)


pedronauck (changes requested):
I found one blocking issue while reviewing recursive task discovery.

In `internal/core/tasks/walker.go`, `taskFileNamesRecursive` currently handles `filepath.WalkDir` errors for non-root directories by logging and returning `filepath.SkipDir`. That silently drops an entire subtree when a directory that may contain `task_NNN.md` files is unreadable or otherwise fails during traversal. For `--recursive`, that means a run can execute an incomplete task set while reporting success.

This is an error-swallowing workaround; discovery should fail loudly for unexpected walk errors. The skip behavior should be reserved for the explicit skip predicate: dot-prefixed dirs, underscore-prefixed dirs, `reviews-*`, `adrs`, and `memory`.

Requested changes:
- Propagate `WalkDir` traversal errors for non-skipped directories, including the relative path in the error context.
- Add a canonical walker test proving an unreadable or non-walkable directory returns an error instead of being skipped.

I verified the fix locally before reverting my local patch because I cannot push to the fork: the focused walker test and lint passed, and full `make verify` passed in the isolated worktree.

pblobp:
Thank you, @pedronauck! I've just addressed your comment.

PR #158, opened 2026-05-17
fix: run managed upgrade commands

MatheusBBarni:
## What changed

`compozy upgrade` now runs the package-manager command for managed installs instead of just printing it. Direct binary installs still use the self-updater.

The managed path now stays tied to the install it detected:
- NPM installs derive the global prefix from `.../lib/node_modules/...` or `.../node_modules/...`, require the package manager in the detected prefix, execute that verified executable path, and set `NPM_CONFIG_PREFIX` before running `npm install -g @compozy/cli@latest`.
- Homebrew installs derive the Homebrew prefix, require `<prefix>/bin/brew`, and execute that verified `brew` path.
- Go installs set `GOBIN` to the bin directory that contains the current `compozy` binary before running `go install github.com/compozy/compozy/cmd/compozy@latest`.

## Why this was broken

The command already knew how the binary was installed. The bug was the last step: for NPM, Homebrew, and Go, it returned after writing the command to stdout. So the CLI looked like it had handled the upgrade, but the installed version never changed.

Review caught two follow-up risks. First, validating a prefix was not enough if `exec.CommandContext` still started from a bare `npm` or `brew`; the code now pins `cmd.Path` to the verified executable. Second, npm detection now handles both Unix-style `lib/node_modules` and Windows/global `node_modules` layouts.

The screenshot shows the original loop: `compozy upgrade` keeps printing the npm command and `compozy --version` stays on 0.2.3 until the npm command is run manually.

compozy-upgrade-not-updating

## Verification

Ran `env -u GOROOT make verify`.

Passed:
- frontend lint, typecheck, tests, and build
- `golangci-lint`: 0 issues
- Go race suite: 3094 tests, 3 skipped
- Playwright daemon UI e2e: 5 passed

## Summary by CodeRabbit

* **Improvements**
  * Upgrade now runs package-manager upgrades (Homebrew, npm, Go) using isolated executables and composed environments instead of only printing commands.
  * Better detection of current install method and safer executable selection across platforms.

* **Documentation**
  * Upgrade help text clarified to state it runs the package-manager command.

* **Tests**
  * Expanded coverage for install detection, managed-command execution, environment composition, Windows specifics, and error propagation; adjusted test parallelism where needed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/compozy/compozy/pull/158?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/pr-assets/compozy-upgrade-not-updating.png
pedronauck, line 1: remove this from git

pedronauck (changes requested):

.github/pr-assets/compozy-upgrade-not-updating.png
MatheusBBarni: removed!

PR #163, opened 2026-05-25
fix: add Windows daemon support

### LINKED ISSUE #148: daemon start fails on Windows: "another process has locked a portion of the file" with no other process holding the lock
## Environment
- OS: Windows (Windows 10/11)
- Compozy version: 0.2.1 (commit=4bdb945, date=2026-05-01)

## Steps to reproduce
1. Fresh install via `irm https://claude.ai/install.ps1 | iex` (note: actually compozy install)
2. Run `compozy daemon start --foreground`

## Expected
Daemon starts.

## Actual
Error: daemon: write daemon lock "C:\\Users\\<user>\\.compozy\\daemon\\daemon.lock": 
write C:\Users\<user>\.compozy\daemon\daemon.lock: 
The process cannot access the file because another process has locked a portion of the file.

## Diagnostics performed
- Verified no other process holds the file with Sysinternals `handle.exe` → "No matching handles found"
- Deleted `~/.compozy/daemon` and `~/.compozy` entirely → error persists on next start
- Added Windows Defender exclusion for `~/.compozy` → no change
- Confirmed user profile is NOT under OneDrive sync (no ReparsePoint on `%USERPROFILE%`)
- No other compozy processes running (`Get-Process *compozy*` returns empty)
- Restarted machine → error persists

This looks like the daemon is opening the lock file with an exclusive byte-range lock 
and then trying to write to it from the same handle in a way that conflicts on Windows.

### LINKED ISSUE #157: daemon start fails on Windows: "Error: wait for daemon readiness: read daemon info: daemon: read daemon info"
When I try to run the `compozy daemon start` command on Powershell:

```
PS D:\code\qr-code> compozy daemon start      
Error: wait for daemon readiness: read daemon info: daemon: read daemon info "C:\\Users\\Fernando Paladini\\.compozy\\daemon\\daemon.json": open C:\Users\Fernando Paladini\.compozy\daemon\daemon.json: The system cannot find the file specified.
```

Also, if I try to run any other command directlt on Powershell, I always got the same error:
```
PS D:\code\qr-code> compozy reviews fix qr-code-microsaas

┌───────────────────────────────────┐
│  COMPOZY // INTERACTIVE INPUT     │
│  Collect parameters for this run  │
└───────────────────────────────────┘
                                     

OK PARAMETERS COLLECTED      
Flags are ready to be applied
Error: wait for daemon readiness: read daemon info: daemon: read daemon info "C:\\Users\\Fernando Paladini\\.compozy\\daemon\\daemon.json": open C:\Users\Fernando Paladini\.compozy\daemon\daemon.json: The system cannot find the file specified.
```

If I run the skills from the VS Code Copilot Chat, it works fine. The same from other development tools. Did I do something wrong while setting up Compozy in my Windows environment?

### LINKED ISSUE #170: Bug: Daemon fails to start on Windows — LockFileEx prevents writeLockPID from writing to the locked file
### Environment

- OS: Windows 11 (corporate laptop, no admin privileges)
- Compozy version: 0.2.4 (installed via `npm install -g @compozy/cli`)
- Architecture: windows-amd64

### Symptoms

Any command that requires the daemon (`tasks run`, `exec`, `daemon start --foreground`) fails with:

```
Error: daemon: write daemon lock "C:\Users\<user>\.compozy\daemon\daemon.lock": write C:\Users\<user>\.compozy\daemon\daemon.lock: The process cannot access the file because another process has locked a portion of the file.
The daemon cannot start on Windows at all, making the entire 0.2.x CLI unusable on this platform.
```

### Root Cause

In `internal/daemon/lock.go`, the `acquireLock` function:

1. Line 75 — Creates a `gofrs/flock` instance and opens the file (`os.OpenFile` with `O_CREATE|O_RDWR`)
2. Line 76 — Calls `fileLock.TryLock()` which invokes Windows `LockFileEx` with `LOCKFILE_EXCLUSIVE_LOCK` on byte range `[0, 1)`
3. Line 89 — Calls `writeLockPID(cleanPath, pid)` which opens a second file handle via `os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o600)`

Step 3 fails because on Windows, `LockFileEx` enforces mandatory byte-range locks across all handles, including handles opened by the same process. The `O_TRUNC` flag in the second `OpenFile` attempts to modify the locked byte range (offset 0), which is denied by the OS.
This is a fundamental behavioral difference between platforms:

- **Linux**: `flock()` is advisory and per-open-file-description — the same process can freely read/write the locked file through any handle.
- **Windows**: `LockFileEx` is mandatory and per-handle — no other handle (even in the same process) can perform I/O on the locked byte range.

### Reproduction (minimal .NET proof)

```
// Simulates exactly what `gofrs/flock` + `writeLockPID` does
var fs = new FileStream(path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.ReadWrite);

var overlapped = new NativeOverlapped();
LockFileEx(fs.SafeFileHandle, LOCKFILE_EXCLUSIVE_LOCK | LOCKFILE_FAIL_IMMEDIATELY, 0, 1, 0, ref overlapped);
// Lock acquired successfully

// Now simulate writeLockPID: open second handle with truncate
var fs2 = new FileStream(path, FileMode.Create, FileAccess.Write, FileShare.ReadWrite);
// ^^^ THROWS: "The process cannot access the file because another process has locked a portion of the file."
```

### Suggested Fix

Option A (minimal change): In `writeLockPID`, instead of opening a new file handle, write through the flock's existing handle. The `*flock.Flock` already holds the open `*os.File` internally. Either:

- Expose a write method on the Lock struct that uses the flock's file handle directly, or
- Pass the flock's file handle to writeLockPID

```
// Instead of opening a new handle:
func writeLockPID(f *os.File, pid int) error {
    if _, err := f.Seek(0, io.SeekStart); err != nil {
        return err
    }
    if err := f.Truncate(0); err != nil {
        return err
    }
    if pid > 0 {
        if _, err := fmt.Fprintf(f, "%d\n", pid); err != nil {
            return err
        }
    }
    return f.Sync()
}
```
**Note**: `Truncate` and `Seek` on the **same handle** that holds the lock are permitted by Windows — only cross-handle I/O is blocked.

**Option B** (alternative): Use flock.Flock's underlying file descriptor (via Fh() method in gofrs/flock v0.13+) to write the PID directly through the lock-owning handle.
**Option C** (fallback): On Windows, skip the PID write after lock acquisition. The lock file itself already guarantees singleton ownership — the PID content is purely informational (used for stale detection, which a successful `TryLock` already implies is unnecessary).

**Affected Code**
File                                          |  Line        |  Issue
`internal/daemon/lock.go`      | 89           |  `writeLockPID` called after `TryLock` with a new handle
`internal/daemon/lock.go`      | 167-168  |  `os.OpenFile` with `O_TRUNC` conflicts with locked byte range

**Impact**
- All Windows users on compozy 0.2.x cannot use any daemon-dependent feature (tasks run, exec, reviews run, etc.)

maxplangg:
Fixes #148
Fixes #157
Fixes #170

## Problem

Three independent failures prevented the daemon from running on Windows — all causing a silent exit with no error shown to the user:

1. **Lock file sharing violation (#148)** — lock uses LockFileEx on Windows, which holds a byte-range lock that blocks *all* file handles, including those in the same process. Writing the PID back to the locked file returned ERROR_SHARING_VIOLATION.
2. **Silent crash on startup (#157)** — syncDir called ile.Sync() on a directory opened read-only. On Windows, FlushFileBuffers requires a writable handle; it returns ERROR_ACCESS_DENIED, causing the daemon to exit before writing daemon.json.
3. **SQLite URI rejected (#157)** — ilepath.ToSlash produces C:/path/to/db on Windows. Passed to url.URL, this serialises as ile://C:/... where the SQLite URI parser interprets C: as a network authority rather than a drive letter, failing to open the database.
4. **Daemon inherited Ctrl+C** — The daemon process inherited the parent console's process group, so any Ctrl+C sent to the CLI also terminated the daemon.

## Changes

| File | Change |
|------|--------|
| internal/daemon/sync_dir_unix.go | Extracted existing syncDir implementation (non-Windows build) |
| internal/daemon/sync_dir_windows.go | No-op syncDir on Windows — NTFS journal guarantees durability after atomic rename |
| internal/daemon/lock_pid_other.go | lockPIDPath identity function for non-Windows (PID stored in lock file) |
| internal/daemon/lock_pid_windows.go | lockPIDPath returns path + .pid sidecar to avoid LockFileEx conflict |
| internal/daemon/lock.go | Use lockPIDPath() helper for all PID read/write calls |
| internal/daemon/info_lock_test.go | Update tests to use lockPIDPath() to reflect platform-correct path |
| internal/daemon/info.go | Remove syncDir (moved to platform-specific files) |
| internal/cli/daemon_launch_windows.go | SysProcAttr with CREATE_NEW_PROCESS_GROUP to detach daemon from parent console |
| internal/cli/daemon_launch_other.go | Add windows to the exclusion build constraint |
| internal/store/sqlite.go | Prepend / to drive-letter paths so url.URL produces valid ile:///C:/... URIs |

## Impact on Unix / Linux / macOS

No behavioral change. All Windows-specific code is behind //go:build windows build tags. The lockPIDPath non-Windows implementation is an identity function (returns the same path). The SQLite fix only triggers when the path does not start with /, which never happens on Unix. The syncDir move is a pure refactor with identical logic.

## Verification

`
make verify
`

- golangci-lint: 0 issues
- Go race suite: all tests pass
- Build: compiles cleanly for Windows target

## Summary by CodeRabbit

* **Chores**
  * Improved cross-platform daemon launching and process detachment (Windows/unix behaviors)
  * Refined lock/PID file handling for more reliable PID tracking and release
  * Enhanced directory sync/durability handling per platform
  * Normalized database file path handling for consistent connection URIs

* **Tests**
  * Expanded and hardened test coverage for daemon lifecycle, locking, SQLite behavior, and store utilities

[Review Change Stack](https://app.coderabbit.ai/change-stack/compozy/compozy/pull/163?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/cli/daemon_launch_other.go
+++ b/internal/cli/daemon_launch_other.go
@@ -1,4 +1,4 @@
-//go:build !(darwin || dragonfly || freebsd || linux || netbsd || openbsd)
+//go:build !(darwin || dragonfly || freebsd || linux || netbsd || openbsd || windows)
 
 package cli
 
--- a/internal/cli/daemon_launch_windows.go
+++ b/internal/cli/daemon_launch_windows.go
@@ -1,0 +1,15 @@
+//go:build windows
+
+package cli
+
+import (
+	"syscall"
+)
+
+func daemonLaunchSysProcAttr() *syscall.SysProcAttr {
+	// CREATE_NEW_PROCESS_GROUP detaches the daemon from the parent console's
+	// process group so that Ctrl+C events are not forwarded to the daemon.
+	return &syscall.SysProcAttr{
+		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
+	}
+}
--- a/internal/daemon/boot_test.go
+++ b/internal/daemon/boot_test.go
@@ -30,7 +30,7 @@ func TestStartRemovesStaleArtifactsAndMarksReady(t *testing.T) {
 	if err := os.WriteFile(paths.SocketPath, []byte("stale"), 0o600); err != nil {
 		t.Fatalf("write stale socket marker: %v", err)
 	}
-	if err := os.WriteFile(paths.LockPath, []byte("999001\n"), 0o600); err != nil {
+	if err := os.WriteFile(lockPIDPath(paths.LockPath), []byte("999001\n"), 0o600); err != nil {
 		t.Fatalf("write stale lock file: %v", err)
 	}
 
@@ -71,7 +71,7 @@ func TestStartRemovesStaleArtifactsAndMarksReady(t *testing.T) {
 	if currentInfo.State != ReadyStateReady {
 		t.Fatalf("current info state = %q, want %q", currentInfo.State, ReadyStateReady)
 	}
-	lockPID, err := readLockPID(paths.LockPath)
+	lockPID, err := readLockPID(lockPIDPath(paths.LockPath))
 	if err != nil {
 		t.Fatalf("readLockPID() error = %v", err)
 	}
@@ -254,7 +254,7 @@ func TestStartCleansUpAfterPrepareFailure(t *testing.T) {
 	if _, statErr := os.Stat(paths.InfoPath); !os.IsNotExist(statErr) {
 		t.Fatalf("expected info file to be removed after prepare failure, stat err = %v", statErr)
 	}
-	pid, readErr := readLockPID(paths.LockPath)
+	pid, readErr := readLockPID(lockPIDPath(paths.LockPath))
 	if readErr != nil {
 		t.Fatalf("readLockPID() error = %v", readErr)
 	}
@@ -336,7 +336,7 @@ func TestStartRebuildsMismatchedRuntimeArtifacts(t *testing.T) {
 	if err := WriteInfo(paths.InfoPath, existingInfo); err != nil {
 		t.Fatalf("WriteInfo(existing) error = %v", err)
 	}
-	if err := os.WriteFile(paths.LockPath, []byte("2222\n"), 0o600); err != nil {
+	if err := os.WriteFile(lockPIDPath(paths.LockPath), []byte("2222\n"), 0o600); err != nil {
 		t.Fatalf("write stale lock file: %v", err)
 	}
 	if err := os.WriteFile(paths.SocketPath, []byte("stale"), 0o600); err != nil {
@@ -369,7 +369,7 @@ func TestStartRebuildsMismatchedRuntimeArtifacts(t *testing.T) {
 		t.Fatalf("current info started_at = %v, want %v", currentInfo.StartedAt, startedAt)
 	}
 
-	lockPID, err := readLockPID(paths.LockPath)
+	lockPID, err := readLockPID(lockPIDPath(paths.LockPath))
 	if err != nil {
 		t.Fatalf("readLockPID() error = %v", err)
 	}
--- a/internal/daemon/info.go
+++ b/internal/daemon/info.go
@@ -134,18 +134,3 @@ func RemoveInfo(path string) error {
 	}
 	return nil
 }
-
-func syncDir(path string) error {
-	dir, err := os.Open(path)
-	if err != nil {
-		return fmt.Errorf("daemon: open directory %q for sync: %w", path, err)
-	}
-	defer func() {
-		_ = dir.Close()
-	}()
-
-	if err := dir.Sync(); err != nil {
-		return fmt.Errorf("daemon: sync directory %q: %w", path, err)
-	}
-	return nil
-}
--- a/internal/daemon/info_lock_test.go
+++ b/internal/daemon/info_lock_test.go
@@ -146,7 +146,7 @@ func TestAcquireLockLifecycle(t *testing.T) {
 		t.Fatalf("lock.StalePID() = %d, want 0", got)
 	}
 
-	pid, err := readLockPID(path)
+	pid, err := readLockPID(lockPIDPath(path))
 	if err != nil {
 		t.Fatalf("readLockPID() error = %v", err)
 	}
@@ -157,7 +157,7 @@ func TestAcquireLockLifecycle(t *testing.T) {
 	if err := lock.Release(); err != nil {
 		t.Fatalf("lock.Release() error = %v", err)
 	}
-	pid, err = readLockPID(path)
+	pid, err = readLockPID(lockPIDPath(path))
 	if err != nil {
 		t.Fatalf("readLockPID(after release) error = %v", err)
 	}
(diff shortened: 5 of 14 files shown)


pedronauck (changes requested):
@maxplangg 

## Findings

### [P2] SQLite Database Path Handling

- **Issue:** Relative SQLite database paths now resolve to filesystem-root paths.
    - In `internal/store/sqlite.go:99`, a leading `/` is unconditionally prepended if the path does not start with `/`. This correctly addresses Windows drive-letter paths, but it also incorrectly modifies valid relative paths on Unix-like systems—converting them into absolute `file:///` URIs, which point to the root instead of remaining relative.
    - `OpenSQLiteDatabase` does **not** require absolute paths. Both `globaldb.Open` and `rundb.Open` accept and use caller-provided paths as-is, so this is a behavioral regression for any consumer using relative database paths.
    - **Recommendation:** The path normalization should be more narrowly targeted. Only add a leading slash when handling Windows drive-letter paths, or convert the input path to an absolute path explicitly/document this contract.
    - **Action item:** Add regression tests covering both `relative.db` and a Windows drive-letter path to ensure both scenarios are correctly handled.

### [P2] Windows PID Sidecar Path Usage in Tests

- **Issue:** The new Windows PID sidecar path is not used in current daemon startup lifecycle tests. As a result, Windows test runs could fail or miss the stale-PID scenario this PR intends to address.
    - The PR introduces `lockPIDPath` for consistent PID file access; on Windows, `internal/daemon/lock_pid_windows.go:10` returns `daemon.lock.pid`.
    - However, existing lifecycle tests (e.g. `internal/daemon/boot_test.go:33`, `boot_test.go:74`, `boot_test.go:339`, `boot_test.go:372`) still assert or manipulate `paths.LockPath` directly.
    - On Windows, `Start` will use `.pid`, while tests interact with `daemon.lock`, so Windows behavior isn't properly validated and could fail if executed.
    - **Recommendation:** Update all relevant tests to use `lockPIDPath(paths.LockPath)` when checking the "persisted daemon PID" invariant.
    - **Action item:** Add a Windows CI/test leg if the PR is intended to support and close Windows daemon-side support.

---

## Merge Recommendation

Not ready to merge. The overall direction is reasonable, and Windows compile checks are passing, but:
- The SQLite URI change is overly broad and regresses relative-path behavior.
- Windows lifecycle test coverage for the new PID path is incomplete for the claimed feature.

Address these issues before merging.

PR #179, opened 2026-05-29
fix: keep multi-run task timers ticking

MatheusBBarni:
## Summary
- keep the multi-run UI clock loop owned by the parent tabbed model
- continue clock ticks even when the active tab has not started its child run yet
- add regression tests for queued-tab attach and queued-tab advance cases

## Verification
- GOROOT="/opt/homebrew/opt/go/libexec" make verify

## Summary by CodeRabbit

* **Refactor**
  * Improved clock tick scheduling reliability in multi-run operations.

* **Tests**
  * Added test coverage for clock tick initialization and continuation behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/compozy/compozy/pull/179?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #180, opened 2026-05-31
release: Release v0.2.8

pedronauck:
## Release v0.2.8

This PR prepares the release of version v0.2.8.

### 🐛 Bug Fixes

- Keep multi-run task timers ticking (#179)

##### Run multiple task workflows from one command
`compozy tasks run` now accepts a `--multiple` flag that enqueues several task workflows through a single daemon-owned parent run, with one TUI tab per requested slug. The single-workflow path (`compozy tasks run <slug>`) is unchanged — `--multiple` is purely additive, so existing habits, scripts, and CI invocations keep working byte-for-byte.

### Starting a multi-run

Pass one comma-separated slug list. The same runtime flags (`--ide`, `--model`, `--stream`, `--detach`, `--attach`, etc.) apply to every workflow in the batch:

```bash
# Enqueue two workflows with shared runtime defaults
compozy tasks run --multiple alpha,beta

# Pick the runtime once for the whole queue
compozy tasks run --multiple alpha,beta --ide codex --model gpt-5.5

# Stream or detach the parent queue
compozy tasks run --multiple alpha,beta --stream
compozy tasks run --multiple alpha,beta --detach
```

Use `--multiple` when one command should drive an ordered batch with identical flags. Keep using `tasks run <slug>` for a single workflow or for scripts that expect exactly one run ID per invocation.

`--multiple` cannot be combined with a positional slug or with `--name`.

### Scheduling mode

Scheduling is controlled by `[tasks.run] run_multiple_mode` in `.compozy/config.toml` or `~/.compozy/config.toml`:

```toml
[tasks.run]
run_multiple_mode = "enqueued"
```

| Mode         | Behavior                                                                                                                       |
| ------------ | ------------------------------------------------------------------------------------------------------------------------------ |
| `"enqueued"` | **Default.** Runs one child workflow at a time, in the requested order. Later slugs stay queued until the active one finishes. |
| `"parallel"` | Accepted for forward-compatible config, but V1 prints a fallback message and still runs the queue enqueued.                    |

When unset, the built-in default is `"enqueued"`. The value is validated at config load: an empty string is rejected, and anything other than `"enqueued"` or `"parallel"` fails with a clear error.

Real parallel multi-run waits on git worktree isolation so concurrent agents never edit the same checkout simultaneously — that lands in V2. Configuring `"parallel"` today is safe: Compozy explains the fallback and proceeds in enqueued order.

### Tabbed TUI

The task-run TUI shows one tab per requested slug:

- Queued tabs appear before their child run exists.
- The running tab shows the familiar single-task surface.
- Completed, failed, and canceled tabs stay available for inspection.

The quit dialog applies to the parent queue:

- **Close TUI** — detaches and leaves the queue running in the daemon.
- **Stop Run** — cancels the parent queue, cancels the active child, and marks queued workflows canceled.
- **Cancel** — returns to the TUI without changing execution.

### Notes

- The multi-run path is daemon-owned end to end: a new parent run orchestrates independent child task runs over the same snapshot-plus-stream transport used by single runs.
- `--dry-run` still previews prompts without executing.


pedronauck:

### 📊 Build Summary
- **Version**: 0.0.0-1780247883
- **Commit**: a55d7ac

### 📦 Built Artifacts
Not available.

---

PR #200, opened 2026-06-15
feat: worktree-backed parallel multi-run for tasks run --multiple

MatheusBBarni:
## What this does

`compozy tasks run --multiple alpha,beta --parallel` now actually runs the tasks at the same time. Until now `parallel` was an accepted value that quietly fell back to sequential (`enqueued`) execution, so the flag was a promise the daemon never kept. This change makes it real: each child task runs in its own git worktree, several at once, bounded by a configurable limit.

`enqueued` stays the default. You only get parallelism when you ask for it, with `--parallel` or `[tasks.run] run_multiple_mode = "parallel"`.

## Why

Running independent task batches one after another wastes time when they don't depend on each other. The point of this work is wall-clock time for solo developers running independent tasks, without losing the per-task status and manual-review handoff the existing multi-run flow already gives you. Each child keeps its worktree after the run so you can read the diffs yourself. Nothing is auto-merged or auto-pushed.

The full background lives in the PRD, tech spec, and ADRs under `.compozy/tasks/multi-task-run/`. ADR-005 (worktree-backed parallel as opt-in) and ADR-008 (configurable limit) are the load-bearing ones for this PR.

## What you get

- `--parallel` and `--parallel-limit N` on `tasks run --multiple`.
- A config key `[tasks.run] run_multiple_parallel_limit` (default `2`, must be positive).
- Limit precedence: the `--parallel-limit` flag wins, then the config key, then the default of 2. Mode precedence: `--parallel`, then `run_multiple_mode`, then `enqueued`.
- Each child runs in a detached worktree under `~/.compozy/state/worktrees/...`, kept after the run.
- Fail-late behavior: if one child fails, the rest keep going. The parent ends `failed` and names which slugs failed, but only after every child has settled. It ends `completed` only when all children complete.
- Cancelling the parent cancels the running children and marks the not-yet-started ones canceled.
- The TUI shows each child's worktree path, branch, and status. The CLI stream tags child events with `worktree=<path>` and prints a final handoff block listing every slug, its status, run id, and worktree.

One thing the README is explicit about: worktrees isolate git working state and nothing else. They do not isolate ports, credentials, caches, provider rate limits, or databases, and they will not prevent merge conflicts. Don't expect full sandboxing.

## How it works

The daemon already owned a sequential multi-run coordinator (`task_multi`). This PR turns it into one mode-aware scheduler that dispatches to either the existing enqueued path (unchanged) or the new parallel path. Both share the same event, snapshot, and cancellation helpers, so there aren't two coordinators drifting apart.

The parallel path does four things:

1. Resolves the parent's branch and HEAD once, and rejects a detached checkout up front. You can't base worktrees on an ambiguous HEAD.
2. Allocates a detached worktree per child at a deterministic path, registers it as its own workspace, and remaps the child runtime onto it.
3. Fans children out with a counting semaphore sized to the resolved limit. Every worker goroutine is owned by the parent and joined before the parent reports a terminal status, so nothing is left running in the background.
4. Aggregates after all workers finish: success only if every child succeeded, otherwise a single error naming the failed slugs.

Worktree metadata (path, base branch, base commit, status) travels on the existing `task.multi.*` item events and gets rebuilt into the snapshot, so attaching late or recovering after a crash still shows where each child ran.

## The SQLite change worth a close look

This is the part to review carefully. The daemon's global catalog is SQLite in WAL mode (modernc driver). Running children at the same time surfaced a real bug: two children starting together both open a deferred transaction, both read, then both try to upgrade to a write. SQLite returns `SQLITE_BUSY` immediately for that read-to-write upgrade, and `busy_timeout` does not retry it. Under the race detector it reproduced every time; without it the timing usually hid the failure.

The fix is `_txlock=immediate` on the DSN, so write transactions start with `BEGIN IMMEDIATE` and take the write lock up front. Concurrent writers then wait on `busy_timeout` (5s) instead of deadlocking. It's a one-line DSN change, but it touches every database opened through `store.OpenSQLiteDatabase`, not just multi-run, so I ran it against the whole suite (3669 tests, race detector on) with no regressions. I also checked the modernc driver source to confirm it honors `_txlock` before depending on it.

I looked at serializing child startup instead, but that only moves the problem: children also write concurrently during execution and teardown. Fixing the lock mode at the source is the honest fix.

## Tests

- Config parsing, merge precedence, defaults, and zero/negative rejection for the new limit.
- Contract and OpenAPI round-trips for the request and snapshot fields, plus old-payload compatibility for the event payloads.
- Daemon scheduler tests for bounded concurrency (limit 2 with 3 children never starts a third before a slot frees, verified with a channel handshake rather than sleeps), fail-late (a failing sibling doesn't stop the others and the parent names the failed slug), cancellation (running child canceled, not-started item canceled, repeated cancel stays consistent), and the resolved limit landing on the started event.
- TUI model tests for the worktree status line, and CLI tests for the handoff block, including that an aggregate failure still exits non-zero with the handoff printed.
- Two end-to-end tests driving the real CLI against an in-process daemon with actual git worktrees.
- A doc-drift guard so the README flags and the event reference stay in sync with the code.

It all passes under `make verify` (fmt, lint at zero issues, race tests, frontend checks, and the Playwright smoke suite).

## Out of scope

No auto-merge, no auto-push, no conflict detection, no dependency-aware scheduling, no automatic worktree cleanup. Those are later phases in the PRD.

One known rough edge, recorded in the workflow memory: a `--dry-run` parallel run still creates real worktrees on disk, because only child execution is short-circuited. A follow-up could plan worktree paths without actually running `git worktree add`.

## Commits

Ten task commits plus one for the planning docs, each gated by a green `make verify`:

| Commit | What it adds |
|---|---|
| task_01 | configurable parallel limit in workspace config |
| task_02 | multi-run API and client contracts |
| task_03 | `--parallel` / `--parallel-limit` CLI controls and request wiring |
| task_04 | worktree metadata on events and snapshot |
| task_05 | git worktree allocation and path planning |
| task_06 | mode-aware `task_multi` scheduler refactor |
| task_07 | register and remap parallel children to worktrees |
| task_08 | bounded fanout, fail-late aggregation, and the SQLite fix |
| task_09 | worktree handoff rendering in the TUI and CLI |
| task_10 | documentation and end-to-end coverage |

## Trying it

```
compozy tasks run --multiple alpha,beta,gamma --parallel --parallel-limit 3
```

Run it from a git workspace where each slug's task files are committed. The worktrees are detached checkouts of HEAD, so the task directories have to exist in the commit.

## Summary by CodeRabbit

* **New Features**
  * Added parallel execution for `compozy tasks run --multiple` via `--parallel`, using git worktree isolation.
  * Added `--parallel-limit` (default: 2) and `run_multiple_parallel_limit` config to cap concurrency; multi-run requests/events/UI now include `parallel_limit` and worktree metadata.
* **Improvements**
  * Parallel mode now consistently forwards and enforces the effective `parallel_limit`, including correct request encoding and snapshot reconstruction.
  * Streamed output and UI tabs now display worktree handoff details with dash placeholders when metadata is missing.
* **Documentation**
  * Updated CLI help text and OpenAPI/event contracts to reflect the new fields.


pedronauck (changes requested):
I reviewed this PR from an isolated worktree and ran three independent review passes (Skeptic, Architect, Minimalist). The core scheduler/worktree implementation looks solid overall, and `make verify` passes, but I’m requesting changes for one public contract mismatch that will affect generated clients.

## Blocking change request

`TaskRunMultipleItem.status` is inconsistent across the live daemon, event docs, OpenAPI, and generated TypeScript client.

The daemon emits and snapshots `"running"` for an active child item:
- `internal/daemon/task_multi.go:25` defines `taskMultiItemStatusRunning = "running"`.
- `internal/daemon/task_multi.go:1121` copies the event payload status directly into the API snapshot item.
- `docs/events.md:370` documents the item status set as `queued`, `running`, `completed`, `failed`, or `canceled`.

But the public schema and generated TS client still advertise `"active"` and omit `"running"`:
- `openapi/compozy-daemon.json:1527` has `["queued", "active", "completed", "failed", "canceled"]`.
- `web/src/generated/compozy-openapi.d.ts:987` mirrors that as `"queued" | "active" | "completed" | "failed" | "canceled"`.
- The new metadata round-trip test in `internal/api/contract/contract_test.go:558` also preserves `Status: "active"`, which reinforces the stale schema value instead of the runtime value.

Failure scenario: while a multi-run child is running, `GET /api/task-runs/multiple/:run_id/snapshot` can return `status: "running"`, which is outside the generated TypeScript union. Consumers generated from this PR’s OpenAPI contract can reject, mishandle, or fail to exhaustively handle a normal live snapshot. `"active"` is effectively dead for this path unless the daemon explicitly remaps it.

Please either:
1. Update the OpenAPI schema/source and generated TypeScript types to use `"running"`, and update the contract tests to prove snapshot item statuses match runtime/event statuses, or
2. Intentionally remap the daemon snapshot status to `"active"` everywhere and update the event docs/runtime tests accordingly.

## Non-blocking notes

- `_txlock=immediate` in `internal/store/sqlite.go` is a process-wide SQLite behavior change, not just a parallel multi-run fix. The rationale makes sense for concurrent child writers, but it is worth calling out in the PR because explicit read transactions elsewhere may now wait for or contend on the writer lock.
- The TUI now reserves a permanent `worktree —` line even for default enqueued multi-runs where worktree metadata is never populated. That is not a merge blocker, but suppressing the line until any worktree metadata exists would preserve transcript space and avoid a confusing placeholder.
- If a journal emit fails immediately after worktree allocation, the worktree can exist without metadata in the snapshot. Narrow window, but worth considering in follow-up cleanup/recovery work.

Verification I ran locally in `../_worktrees/pr-200-review`:
- `make verify` -> exit 0
- Focused checks: `go test ./internal/daemon ...`, `go test ./internal/cli ...`, `go test ./internal/api/contract`, `go test ./internal/api/client`, `go test ./internal/api/core`, narrow `internal/api/httpapi` OpenAPI contract test, `go test ./pkg/compozy/events/...`, and `go test ./internal/core/workspace ./internal/config` all passed.

Issue #32, opened 2026-04-03
✨ Feedbacks & Suggestions


pedronauck:
## Share your feedback and suggestions

This is an open space for the community to share **feedback**, **ideas**, and **feature suggestions** for Compozy.

We'd love to hear from you:
- What do you like or dislike about the current experience?
- What features or improvements would you like to see?
- Any workflow ideas that could make Compozy more useful?

### How to contribute

Simply leave a comment below with your thoughts. Be as detailed as you'd like — context helps us prioritize better.

### Bug reports don't belong here

If you've found a **bug**, **error**, or **unexpected behavior**, please **do not post it here**. Instead, [open a new issue](../../issues/new) with steps to reproduce the problem. Bug reports mixed into this thread will be harder to track and may get lost.

**TL;DR:**
- Feedback, ideas, suggestions → comment here
- Bugs, errors, crashes → [open a separate issue](../../issues/new)

rodrigobranas:
I suggest removing the step-by-step approval for the PRD and Tech Spec. Instead, generate the full preview first and ask for changes at the end. I think most people will naturally click “Next, Next, Finish” and only review everything afterward.

camopy:
> I suggest removing the step-by-step approval for the PRD and Tech Spec. Instead, generate the full preview first and ask for changes at the end. I think most people will naturally click “Next, Next, Finish” and only review everything afterward.

Maybe make this optional. This was one of the things I noticed I enjoyed more compared to the previous Compozy

zamorano-dev:
> I suggest removing the step-by-step approval for the PRD and Tech Spec. Instead, generate the full preview first and ask for changes at the end. I think most people will naturally click “Next, Next, Finish” and only review everything afterward.

I kind of liked that new way of confirmation. So I agree with @camopy in making it optional being the best solution

brunomguimaraes:
I'd like to have some sort of feedback when at the end of a lifecycle. Simple sound could work.
I've drafted something [here](https://github.com/compozy/compozy/pull/96)

iagowp-mthr:
The link to create a new issue is broken

antoniofulg:
EDIT: I noticed the .compozy folder should be in the .gitignore, right? This suggestion doesn't make sense.

What about supporting a counter code prefix for the tasks? I manually added a number as a prefix for the folders, and the IA starts working. But it would be simpler if we could refer to the tasks/reviews by the task/ticket number in the config file:

```toml
[defaults]
[...]

PR #216, opened 2026-06-26
feat: support COMPOZY_HOME env override for home directory

rodrigo123antunes:
## Summary

Adds a `COMPOZY_HOME` environment variable as an opt-in override for the Compozy home root. When set to a non-empty value, `ResolveHomeDir` resolves against that path instead of the implicit `$HOME/.compozy`, and every home-scoped consumer is routed through it so the override is honored consistently: `ResolveHomePaths` / `EnsureHomeLayout`, daemon startup, global config loading, global workspace-marker detection, extension discovery + enablement, and global reusable-agent discovery.

## Why

The daemon today is a singleton per `$HOME`: every workspace that talks to `~/.compozy/daemon/daemon.sock` serializes through one engine. As raised in #32 (comment from @rjsoux), users running multiple independent projects in parallel hit the same socket from every repo and have no first-class way to isolate them. Today the workaround is a "mirror home" that symlinks everything except `.compozy`, which is fragile and unsupported.

This PR delivers **Tier 1** of the request: an official, documented escape hatch. Operators point one shell at one `COMPOZY_HOME`, one at another, and each gets its own daemon, socket, lock, state, and global DB. The CLI `--home` flag and Tier 2 (true parallel runs across workspaces inside one daemon) are intentionally out of scope and can layer on top.

## What Changed

- `internal/config/home.go`
  - New const `HomeEnvVar = "COMPOZY_HOME"`.
  - New package-level `osLookupEnv` stub var, mirroring the existing `osUserHomeDir` pattern, so tests can override env resolution without touching the process.
  - New `lookupHomeOverride()` helper that returns `(value, true)` only when the variable is set to a non-empty, whitespace-trimmed string.
  - `ResolveHomeDir` precedence is now: `COMPOZY_HOME` (if set) → `$HOME/.compozy` (existing behavior).
  - `~` and `~/` prefixes inside `COMPOZY_HOME` are expanded against the current user's home via the existing `ResolvePath` / `expandUserPath` helpers, so `COMPOZY_HOME=~/alt-compozy` works.
- **Routed the remaining home-scoped consumers through `ResolveHomeDir` / `ResolveHomePaths`.** These previously rebuilt `$HOME/.compozy` from `os.UserHomeDir` directly, so a set `COMPOZY_HOME` split-brained — the daemon isolated but these consumers stayed on `$HOME/.compozy`:
  - `internal/core/workspace/config.go` — `resolveConfigPaths` (global config path) and `discoverGlobalWorkspaceMarkerDir` (global workspace-marker detection).
  - `internal/core/extension/discovery.go` + `internal/core/extension/enablement.go` — the enablement store and discovery now root at the Compozy home (the `.compozy` / `COMPOZY_HOME` root) instead of appending `.compozy` to the OS home.
  - `internal/cli/extension/root.go` + `internal/cli/extension/install.go` + `internal/cli/setup_assets.go` — the `ext` flow now separates the OS user home (used only for `setup.ResolverOptions`: `.claude` / `.codex` / `.config`) from the Compozy home (install target, enablement store, discovery), resolving the latter via `ResolveHomeDir`.
  - `internal/core/agents/agents.go` — the global reusable-agent root now honors the override.
- `internal/config/home_test.go` and the touched packages' tests
  - Coverage for: env override used, whitespace trimmed, empty ignored, whitespace-only ignored, `~` expansion, full `ResolveHomePaths` propagation, unset fallback.
  - Tests that touch env vars drop `t.Parallel()` (Go disallows combining it with `t.Setenv`); existing tests keep their parallelism.
  - Broadened `COMPOZY_HOME` guards across the affected packages' tests so an ambient `COMPOZY_HOME` in a developer's environment cannot leak into fixtures.

## Behavior Notes

- **Backward compatible.** No env var set → identical behavior to `main` (resolved against `os.UserHomeDir` + `DirName`).
- **Empty / whitespace-only** `COMPOZY_HOME` is treated as unset. No silent surprises if a script exports `COMPOZY_HOME=""`.
- **Relative paths** in `COMPOZY_HOME` are made absolute via `filepath.Abs`, matching the rest of `ResolvePath`.
- **Cross-platform.** Works on Linux, macOS, and Windows — env-var only, no shell expansion required.
- **Consistent isolation.** The daemon socket/lock/info/log/runs/state/global DB, global config, global workspace-marker detection, extension discovery + enablement, and global reusable agents all resolve against the same root, so two daemons with different `COMPOZY_HOME` values do not collide and there is no split-brain between the daemon and config/extensions.
- **OS home is left untouched.** Legitimate OS-home lookups — `.claude`, `.codex`, XDG (`~/.config/compozy/state.yml`), and path display — continue to use `os.UserHomeDir` and are intentionally unaffected by `COMPOZY_HOME`.
- **No CLI flag in this PR.** The environment variable is the contract. A `--home` flag can be added later as a thin wrapper around the same lookup.

## Testing

- `go test -race ./...` — green for the touched packages (`config`, `core/workspace`, `core/extension`, `core/agents`, `cli`, `cli/extension`, `daemon`).
- Several consumers previously rebuilt `$HOME/.compozy` via `os.UserHomeDir` and were explicitly routed through `ResolveHomeDir` / `ResolveHomePaths` so they honor the override consistently.

## Verification

```text
make fmt                          # clean
make lint                         # 0 issues
make go-build                     # clean
make test                         # green for the touched packages (-race)
```

## Related

- Addresses Tier 1 of the request in #32 (comment by @rjsoux).
- Crosses paths with the planned `parallel` multi-run mode (Tier 2 of the same request); orthogonal change.

## Checklist

- [x] `make fmt` clean
- [x] `make lint` clean (0 issues)
- [x] `make test` passes for the touched packages
- [x] Backward compatible (no env var → unchanged behavior)
- [x] Cross-platform (Linux/macOS/Windows)

## Summary by CodeRabbit

* **New Features**
  * You can now override the Compozy home directory with an environment variable (supports trimming whitespace and `~` path expansion).
* **Bug Fixes**
  * Empty or whitespace-only overrides are ignored, and the app falls back to the standard user home.
  * Compozy-home–derived paths (including derived subpaths) now consistently follow the override when present.
* **Improvements**
  * Extension discovery, enablement state storage, and related workspace/global configuration are now rooted in the Compozy home directory for more consistent behavior.

### DIFF
--- a/internal/cli/extension/display_test.go
+++ b/internal/cli/extension/display_test.go
@@ -678,6 +678,7 @@ func TestPathExistsAndRenderHooksHelpers(t *testing.T) {
 func TestDefaultCommandDepsAndHelperFunctions(t *testing.T) {
 	deps := defaultCommandDeps()
 	if deps.resolveHomeDir == nil ||
+		deps.resolveCompozyHome == nil ||
 		deps.resolveWorkspaceRoot == nil ||
 		deps.isInteractive == nil ||
 		deps.confirmInstall == nil ||
@@ -993,6 +994,7 @@ func newTestDeps(t *testing.T) testDeps {
 
 	deps := defaultCommandDeps()
 	deps.resolveHomeDir = func() (string, error) { return homeDir, nil }
+	deps.resolveCompozyHome = func() (string, error) { return filepath.Join(homeDir, ".compozy"), nil }
 	deps.resolveWorkspaceRoot = func(context.Context) (string, error) { return workspaceRoot, nil }
 	deps.isInteractive = func() bool { return false }
 	deps.confirmInstall = func(*cobra.Command, installPrompt) (bool, error) { return true, nil }
@@ -1127,7 +1129,9 @@ func writeTestFile(t *testing.T, path string, content string) {
 func mustEnablementStore(t *testing.T, homeDir string) *extensions.EnablementStore {
 	t.Helper()
 
-	store, err := extensions.NewEnablementStore(context.Background(), homeDir)
+	// homeDir is the OS-home temp; the store roots at the Compozy home, matching
+	// how the CLI resolves it via resolveCompozyHome (config.ResolveHomeDir).
+	store, err := extensions.NewEnablementStore(context.Background(), filepath.Join(homeDir, ".compozy"))
 	if err != nil {
 		t.Fatalf("create enablement store: %v", err)
 	}
--- a/internal/cli/extension/install.go
+++ b/internal/cli/extension/install.go
@@ -127,7 +127,7 @@ func runUninstallCommand(cmd *cobra.Command, deps commandDeps, rawName string) e
 		return err
 	}
 
-	userPath := filepath.Join(userExtensionsRoot(env.homeDir), name)
+	userPath := filepath.Join(userExtensionsRoot(env.compozyHome), name)
 	exists, err := deps.pathExists(userPath)
 	if err != nil {
 		return fmt.Errorf("inspect user extension path %q: %w", userPath, err)
@@ -254,7 +254,7 @@ func prepareInstallRequest(
 		)
 	}
 
-	installPath := filepath.Join(userExtensionsRoot(env.homeDir), manifest.Extension.Name)
+	installPath := filepath.Join(userExtensionsRoot(env.compozyHome), manifest.Extension.Name)
 	if sameInstallPath(resolvedSource.SourcePath, installPath) {
 		return nil, "", installPrompt{}, fmt.Errorf(
 			"extension %q is already installed at %s",
@@ -402,8 +402,8 @@ func ensureInstallTargetAvailable(deps commandDeps, installPath string, name str
 	return nil
 }
 
-func userExtensionsRoot(homeDir string) string {
-	return filepath.Join(homeDir, ".compozy", "extensions")
+func userExtensionsRoot(compozyHome string) string {
+	return filepath.Join(compozyHome, "extensions")
 }
 
 func sameInstallPath(left string, right string) bool {
--- a/internal/cli/extension/root.go
+++ b/internal/cli/extension/root.go
@@ -12,6 +12,7 @@ import (
 
 	huh "charm.land/huh/v2"
 
+	compozyconfig "github.com/compozy/compozy/internal/config"
 	extensions "github.com/compozy/compozy/internal/core/extension"
 	"github.com/compozy/compozy/internal/core/kernel"
 	"github.com/compozy/compozy/internal/core/workspace"
@@ -27,6 +28,7 @@ type discoverRequest struct {
 
 type commandDeps struct {
 	resolveHomeDir       func() (string, error)
+	resolveCompozyHome   func() (string, error)
 	resolveWorkspaceRoot func(context.Context) (string, error)
 	isInteractive        func() bool
 	confirmInstall       func(*cobra.Command, installPrompt) (bool, error)
@@ -42,7 +44,12 @@ type commandDeps struct {
 }
 
 type commandEnv struct {
-	homeDir       string
+	// homeDir is the OS user home, used only for setup.ResolverOptions
+	// (.claude/.codex/.config detection).
+	homeDir string
+	// compozyHome is the Compozy home root (".compozy" or COMPOZY_HOME), used for
+	// extension install, enablement state, and discovery.
+	compozyHome   string
 	workspaceRoot string
 	store         *extensions.EnablementStore
 }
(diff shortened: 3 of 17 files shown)


rodrigo123antunes:
@pedronauck review this

pedronauck (changes requested):
## Review summary

Thanks for this — the `home.go` implementation itself is clean and correct (clear precedence, safe empty/whitespace fallback, `~` expansion, and the `osLookupEnv` stub mirrors the existing pattern). The **daemon isolation** that motivated #32 (socket/lock/db/runs/state) works as intended.

My concern is **scope, not the new code**: the PR body states that "every caller that flows through it — ... global config ... extension enablement, workspace config" inherits the override. That claim is **not true for three consumers**, because they build the `.compozy` path independently and never go through `ResolveHomeDir`. The result is a split-brain when `COMPOZY_HOME` is set (which is the whole point of the feature): the daemon isolates, but config / extensions / workspace detection stay on `$HOME/.compozy`.

### Findings (most severe first)

**1. `internal/core/workspace/config.go:257` — global/workspace config does NOT respect `COMPOZY_HOME`.**
`resolveConfigPaths` calls `osUserHomeDir()` directly (line 248) and then `ResolveHomePathsFrom(filepath.Join(resolvedHomeDir, DirName))` — the explicit-base variant, which does **not** call `ResolveHomeDir`. So with `COMPOZY_HOME=/foo`, the daemon creates DB/socket under `/foo`, but `config.toml` is still read from `$HOME/.compozy/config.toml`. Daemon and config diverge — directly contradicts the "global config / workspace config" claim.

**2. `internal/core/extension/discovery.go:167` (and `enablement.go:174-178`) — extensions and enablement state do NOT respect `COMPOZY_HOME`.**
`Discovery{WorkspaceRoot: ...}.Discover` calls `NewEnablementStore(ctx, "")`, which falls back to `os.UserHomeDir()`, and the scan joins `".compozy", "extensions"` (hardcoded). Two daemons with different `COMPOZY_HOME` values share the same user extensions and the same `.compozy-state.json` enablement file — the opposite of the isolation promised under "extension enablement".

**3. `internal/core/workspace/config.go:153` — `discoverGlobalWorkspaceMarkerDir` ignores the override.**
Builds `filepath.Join(resolvedHomeDir, model.WorkflowRootDirName)` from `osUserHomeDir()` directly (line 144). Same root cause as #1: global workspace marker detection stays on `$HOME/.compozy` even when the override is set.

**4. (minor) PR body / `internal/update/state.go` — inaccurate "update state" claim.**
The PR lists "update state" as covered, but it lives under `~/.config/compozy/state.yml` (XDG), never under `.compozy`. Not a regression, but the claim is wrong.

### Requested changes

- **Complete the feature** — route `resolveConfigPaths`, `discoverGlobalWorkspaceMarkerDir`, and the extension `homeDir` through `ResolveHomeDir` so the override is honored consistently. The current fix is too shallow (only `home.go`); generalizing the underlying mechanism is preferable to leaving a partial feature.

### Another Change

Tests in other packages (`internal/core/sync`, `internal/daemon`, etc.) that call `ResolveHomePaths` may become flaky if a developer has `COMPOZY_HOME` set in their environment — the `t.Setenv(HomeEnvVar, "")` guard was only added to `home_test.go`. Worth considering a broader guard.

rodrigo123antunes:
Thanks for digging into this @pedronauck, the scope critique was fair. The first pass only touched `home.go`, so the override worked for the daemon but not for the consumers that build the `.compozy` path on their own. I've now routed all of them through `ResolveHomeDir` / `ResolveHomePaths` so `COMPOZY_HOME` is honored everywhere.

Going through your findings:

On (1), `resolveConfigPaths` no longer calls `os.UserHomeDir` and reassembles the path by hand. It uses `ResolveHomePaths()` directly. I kept `globalRoot` as `filepath.Dir(homePaths.HomeDir)` so relative `add_dirs` still resolve against the same base as before in the default layout, and against something sensible under the override. Global config and the daemon now share one root.

On (2), the enablement store and discovery now treat their home field as the Compozy root (the `.compozy` or `COMPOZY_HOME` directory) instead of the OS home with `.compozy` hardcoded, and an empty value falls back to `ResolveHomeDir`. The tricky part was the CLI `ext` flow, where the OS home is genuinely needed for `.claude` / `.codex` / `.config` through `setup.ResolverOptions`. So I kept the two homes separate: OS home stays for setup, and a new `resolveCompozyHome` dep drives install, the store, and discovery. Two daemons with different `COMPOZY_HOME` values now get their own extensions and `.compozy-state.json`.

On (3), `discoverGlobalWorkspaceMarkerDir` just calls `ResolveHomeDir()` now.

On (4), good catch, that claim was wrong. Update state lives under XDG (`~/.config/compozy/state.yml`), never `.compozy`. I removed it from the description along with the "no other call sites needed updates" line, which obviously wasn't true.

While I was at it I found one more consumer in the same boat: `globalAgentsRoot` in `internal/core/agents` was also building `$HOME/.compozy/agents` from `os.UserHomeDir`, so global reusable agents ignored the override too. Routed that one as well.

On your test note, you were right that the guard was too narrow. I broadened it beyond `home_test.go`: `isolateWorkspaceConfigHome` now clears any ambient `COMPOZY_HOME`, and the extension and daemon test stores root at the Compozy home so they line up with how the runtime resolves it.

`make fmt`, `make lint` (0 issues), `make go-build` and `make test` are green for the touched packages (config, core/workspace, core/extension, core/agents, cli, cli/extension, daemon). Left the legitimate OS home lookups (`.claude`, `.codex`, XDG, path display) alone.

PR #226, opened 2026-07-07
fix: correct Kiro CLI ACP model handling

edudutra:
## Summary

Every `compozy exec --ide kiro` run failed. Two root causes:

1. **Unsupported ACP method.** After creating a session, Compozy called
   `session/set_config_option` to switch the model, but `kiro-cli` does not
   implement that method and rejects it with `-32601 Method not found`.
   Kiro receives its model via the `--model` launch flag, so its spec is now
   marked `UsesBootstrapModel` and `SetSessionModel` is a no-op for
   bootstrap-model runtimes (which already pin the model at launch).
2. **Invalid default model.** `DefaultKiroModel` was `anthropic/claude-opus-4-6`,
   which is not a valid Kiro model ID. Kiro exposes IDs like `claude-opus-4.6`
   (see `kiro-cli chat --list-models`), so the default is corrected.

## Changes
- `internal/core/agent/client.go`: skip `session/set_config_option` for `UsesBootstrapModel` runtimes.
- `internal/core/agent/registry_specs.go`: mark Kiro `UsesBootstrapModel: true`.
- `internal/core/model/constants.go`: `DefaultKiroModel` -> `claude-opus-4.6`.
- `internal/core/agent/client_test.go`: regression tests for both behaviors.

## Test plan
- `make verify`: gofmt, golangci-lint (0 issues), all Go tests (`-race`), and build pass.
- Frontend lint/typecheck/unit tests pass.
- Manually verified `compozy exec "..." --ide kiro` now completes a turn.

> Note: the Playwright browser E2E could not run in my local environment
> (Playwright has no Chromium build for Ubuntu 26.04); CI runs it on a
> supported image. The change is backend-only and does not touch the web UI.

## Summary by CodeRabbit

* **New Features**
  * None.

* **Bug Fixes**
  * Improved session handling for runtimes that don’t support switching models during a session by skipping unsupported configuration updates.
  * Better error reporting when model switching configuration is rejected by the runtime.
  * Updated default model selection to a newer identifier.

* **Documentation**
  * Added clarifying inline notes for the bootstrap-model runtime behavior.


pedronauck (changes requested):
The Kiro fix itself is still relevant, but this head is not merge-ready. A merge-tree against current `main` reports content conflicts in `internal/core/agent/client.go` and `internal/core/agent/client_test.go`, and GitHub currently marks the PR `DIRTY`/`CONFLICTING`. The current SHA also has no `verify` CI check.

Please update the branch from current `main`, resolve both conflicts while preserving the newer session/client lifecycle behavior on `main` and this PR's bootstrap-model guard/regression coverage, then run the full `make verify` gate and push the resulting clean head for review.

edudutra:
@pedronauck You were right — there were real conflicts against current `main`. I've rebased onto `main` (v0.2.14); the branch is now a single commit that fast-forwards onto `main` with no conflicts (`mergeable: true`).

**How the conflicts were resolved**

`internal/core/agent/client.go` and `client_test.go` were resolved in favor of `main`. The ACP refactor that landed on `main` (codex/acp integration + `fix: parallel tasks` #231) already replaced the old `SetSessionModel` path with `configureSession` / `configureSessionModel`, where `UsesBootstrapModel` already short-circuits **before** `setSessionConfigValue`. This PR's original guard lived inside `SetSessionModel` (now deleted) and its tests targeted that same method, so keeping them would have been dead/duplicate code — I dropped them rather than reintroduce the old path.

**What this PR still contributes** (the part `main` doesn't have yet — 2 files, +7/−3):

- `registry_specs.go`: mark the Kiro spec `UsesBootstrapModel: true`. Without it, `configureSessionModel` falls through to `setSessionConfigValue` for Kiro (`requiresAdvertisedSessionModel` only covers Cursor/Codex), which `kiro-cli` rejects with `-32601`.
- `constants.go`: `DefaultKiroModel` `anthropic/claude-opus-4-6` → `claude-opus-4.6` (valid Kiro model ID).

The bootstrap-model regression coverage you asked to preserve already exists on `main` (the test helpers default `UsesBootstrapModel: true`, and both bootstrap/non-bootstrap session-setup paths are exercised).

**Verification**

`internal/core/agent` and `internal/core/model` pass with `-race`; `fmt` / `lint` / `build` are clean.

Heads-up on the full gate: `make verify` currently trips on two **pre-existing** `internal/daemon` tests that also fail on `main` in my environment (a parallel `HOME`-env race in `TestParallelTaskChildFallbackOutcomeTreatsCrashAsRecoverable`, and the git-worktree cleanup assertion in `TestRunManagerStartTaskWorktreeChildCleansUpAllocatedWorktreeOnLaunchFailure`). This branch is byte-identical to `main` under `internal/daemon`, so those are unrelated to this change — happy to file them separately if useful.

Ready for re-review.

pedronauck (changes requested):
The branch update resolved the prior merge conflicts, but head `7703a51152399639d3c19ebd81a7731b7fef5d47` is still not merge-ready. The rebase reduced the PR to the Kiro spec flag and default constant, while the Kiro-specific regression coverage described by the PR disappeared. Existing `TestAgentRegistryEntries` launches Kiro only with `resolveModel(spec, "")` and compares against the same `DefaultKiroModel` symbol, so it still passes if `UsesBootstrapModel` regresses to false and does not independently prove the corrected model ID.

Please add a Kiro-owned regression that asserts the spec uses bootstrap model selection, an explicit non-default requested model reaches `kiro-cli acp --model <requested>` at launch without a runtime session model update, and the default resolves to the supported Kiro identifier. Then rerun the full `make verify` gate after the base `goconst` fix lands; the current required `verify` check is red on that inherited failure.

pedronauck (approved):
Added owning-layer regressions for Kiro bootstrap launch selection and the absence of runtime model reconfiguration. The PR merge ref tree exactly matches verified candidate dbea771. Official run 29543373473 passed formatting, golangci-lint with 0 issues, 4,802 Go tests (4 intentional skips), build, extension-module verification, and 5 Playwright E2E tests.

pedronauck (approved):
Final current-main merge tree `ef241977` has tree `dc4c95a830b7f498edf953ea751e97219d32f26a`, exactly matching combined candidate `70916b2`. Official run 29543671867 passed formatting, golangci-lint with 0 issues, 4,875 Go tests (4 intentional skips), build, extension-module verification, and 5 Playwright E2E tests.

PR #230, opened 2026-07-09
fix: recover stalled and wedged multi-runs

MatheusBBarni:
## The problem

A multi-run could freeze partway through and never come back. When an agent stopped emitting output, the attempt sat idle until some external timeout, and in a parallel run a wedged child would hold the parent's join open with no way out. Nothing told you a job had stalled or that one needed a look.

## What changed

The branch adds stall detection and recovery to the run lifecycle.

A watchdog looks for progress on a dedicated idle window. Any ACP session update resets its clock, so it only fires on an attempt that has genuinely gone quiet. When it fires, the job gets one retry from a clean worktree. If it stalls again, the job parks instead of failing: its worktree and journal are kept so you can triage what happened.

Parallel runs get a second layer. Each child's durable journal high-water sequence is tracked, and if it stops advancing for the whole budget, the parent reaps the child through the normal cancel path so the join proceeds. The backstop budget sits strictly above the idle window, so the in-attempt watchdog always gets first chance to heal.

Two supporting fixes make recovery reliable. Terminal commands are now cancellable, wall-clock capped, and parented to the attempt, so a stuck command can't orphan a process. Critical ACP session updates are no longer dropped under backpressure, and stall diagnostics are logged when the watchdog trips.

## Visibility

There's a new parked run status plus job stall and park events. The run UI shows stalled, retrying, and parked states, and prints an end-of-run recovery summary with completed, recovered, and parked counts. When a job parks, an alert fires with the triage detail: why it parked, the last tool call, and the worktree path.

## Config

New stall-detection knobs (idle timeout, child timeout, terminal cap, retries), on by default. An explicit disable turns the whole thing off.

## Testing

`go build` and `go test -race` pass on `internal/core/run/executor` and `internal/daemon`, and golangci-lint is clean on both. The last commit folds in three review fixes: the stall-retry attempt budget (it was reporting an impossible `3/2` when a stall retry preceded ordinary retries), a duplicate alert on parked runs, and a child-policy defaulting gap that could silently disarm the backstop.

## Summary by CodeRabbit

* **New Features**
  * Added “job stalled” and “job parked” states with richer status, progress, and triage details in the CLI and UI.
  * Added configurable parked alerts/sound plus stall-policy support and multi-run recovery summaries (recovered/parked counts).
  * Added terminal command wall-clock timeouts and improved terminal teardown behavior.

* **Bug Fixes**
  * Ensured parked outcomes are treated as terminal and counted separately from failed/canceled runs.
  * Improved session publishing so critical updates never drop under backpressure.

* **Tests**
  * Expanded coverage for stalled/parked rendering, sound selection, run recovery counts, and stall/timeout behavior.


pedronauck (changes requested):
Validated at head `2cb26b3524ee505ea97bb68d040f7c9ea8c24d4c` with the full `make verify` gate (exit 0), but there is a blocking runtime-override contract bug.

`runtimeOverrideInput.Stall` is decoded from JSON with `DisallowUnknownFields`, while `workspace.StallOverrides` only defines `toml` tags. As a result, the API-style payload `{"stall":{"enabled":true,"timeout":"200ms","child_timeout":"500ms","terminal_command_timeout":"1s","retries":1}}` fails with `json: unknown field "child_timeout"` instead of applying the policy. The integration helper in `internal/daemon/child_backstop_test.go` currently hides this by sending Go/PascalCase keys (`Enabled`, `Timeout`, `ChildTimeout`).

Please add the corresponding `json` tags to every `StallOverrides` field, update the integration payload to use the public snake_case shape, and add a parser-level regression proving all stall fields decode and reach `RuntimeConfig` through `parseRuntimeOverrides`/`applyRuntimeOverrideInput`. Then rerun `make verify`.

MatheusBBarni:
@pedronauck Done! Can you review again?

pedronauck (approved):
Approved at head e695fdc3739024ea90290bd24470f3061632ca9f. The snake_case stall override contract is now covered through the public parser/application path, all active review threads are resolved, the focused worktree cleanup regression passed 5/5 with the race detector, and CI run 29432190781 executed the exact current merge ref 29f6303b1516e1bc07802be63f85f70c4bac917c with make verify fully green: lint reported 0 issues, 4,769 tests passed, build completed, and Playwright E2E passed.

antoniofulg:
## Follow-up hardening (2026-07-22)

[PR #242](https://github.com/compozy/compozy/pull/242), in commit [`76e6710`](https://github.com/compozy/compozy/commit/76e6710), closes operational gaps found after this PR merged:

- Concurrent review-fix jobs shared one workspace, so the clean-reset safety guard rejected the first stall with `workspace is shared with sibling jobs` and parked immediately instead of consuming the configured retry. Review batches now receive disposable isolated worktrees and integrate successful deltas back serially.
- Silent terminal commands were counted as agent idleness. A live ACP terminal now suspends the idle timer while retaining the independent absolute terminal cap.
- `job.stalled` and `job.parked` are now projected durably instead of leaving the stored job state at `started`.
- The existing global stall knobs are documented, and `[fix_reviews.stall]` adds per-review overrides for users who prefer spending more tokens on faster clean retries.

The tradeoff is explicit: shorter idle windows and more retries can increase model-token usage. Concurrent review isolation also requires a committed Git baseline and rejects source changes outside the workflow artifact scope; conflicting or parked worktrees are preserved for triage.

PR #234, opened 2026-07-10
fix: make cy-qa-workflow extension work on compozy.tasks/v2 runs

antoniofulg:
## Summary

The `cy-qa-workflow` extension aborts **every** `compozy tasks run` on a `compozy.tasks/v2` workflow. This PR fixes the two independent root-cause defects behind it. A follow-up PR adds a configurable QA runtime on top of this.

## Why

With `cy-qa-workflow` enabled, launching any v2 (frontmatter-graph) task workflow fails during plan discovery, before `run.started`, and no task jobs are created. Because the extension's `plan.pre_discover` hook is `required = true`, its failure aborts the whole run. Each aborted run also leaves a partial artifact that can flip the daemon to `degraded`. The only workaround today is to disable the extension entirely, which loses QA-task automation.

There are two separate causes:

1. The extension ships without a `go.mod`, so its `go run .` subprocess can't even start once installed standalone.
2. Even with the subprocess running, its `host.tasks.create` call is rejected by the daemon, because that host method only understands the **legacy Markdown-table** `_tasks.md`, not the **v2 frontmatter-graph** format that `cy-create-tasks` generates and the runner requires.

### Defect 1 — extension had no Go module

The installed extension contained only `extension.toml` + `main.go`, but `extension.toml` runs the subprocess as `go run .` and `main.go` imports `github.com/compozy/compozy/sdk/extension`. With no `go.mod`, the subprocess failed immediately with `go: go.mod file not found`.

- **`extensions/cy-qa-workflow/go.mod`** (new) — makes the extension the standalone module the docs/scaffolder prescribe. Module path `example.com/cy-qa-workflow`: a deliberately non-publishable path (`example.com` is RFC 2606 reserved), matching what `create-extension` emits (`defaultModuleName → example.com/<name>`) and `docs/extensibility/hello-world-go.md` (`go mod init example.com/hello-go`). The module is an app run via `go run .`, never imported, so the path is only an identifier.
- **`extensions/cy-qa-workflow/go.sum`** (new) — generated by `go mod tidy`; pins the SDK + transitive deps.

### Defect 2 — `host.tasks.create` wasn't `compozy.tasks/v2`-aware

`host.tasks.create` searched for the legacy header `| # | Title | Status | Complexity | Dependencies |` and returned **"tasks table header not found"** on a v2 manifest, aborting the run at `extension.ready`. It now detects the manifest format and mutates the v2 graph instead.

- **`internal/core/tasks/manifest_write.go`** (new) — three helpers on the manifest type:
  - `IsTaskGraphManifestContent` — detects a v2 manifest (schema version or `graph.nodes`).
  - `AppendTaskGraphNode` — adds a node to `graph.nodes` and dependency edges to `graph.edges` (`dep → new`), deduplicated; errors on a dangling dependency so an invalid manifest is never written; re-serializes the frontmatter.
  - `RenderTaskGraphTaskFile` — renders a v2 task file with only task-owned front matter (no `dependencies` key, which the v2 validator rejects on task files).
- **`internal/core/extension/host_writes.go`** — `CreateTask` is now format-aware: it reads `_tasks.md` once (`loadTaskIndexState`) and branches the task-file rendering (`renderHostTaskFile`) and the index update (`buildTaskIndexUpdate`). For a v2 manifest it renders without `dependencies` and appends node + edges; the legacy Markdown-table path is unchanged.

### Build coverage

Adding a `go.mod` makes the extension a **nested module**, which Go excludes from the root `./...`, so `make verify` would silently stop covering it.

- **`Makefile`** — new `verify-extensions` target (build `-o /dev/null`, `go test -race`, `golangci-lint run` per module listed in `EXTENSION_GO_MODULES`), wired into `verify`.
- **`.github/workflows/ci.yml`** — `extensions/**` added to the CI `changes` paths filter so extension-only changes still trigger `verify`.
- **`.gitignore`** — ignore `go.work`/`go.work.sum` (local dev) and the compiled `cy-qa-workflow` binary from `go build .`.
- **`test/frontend-verification-contract.test.ts`** — the `verify` target-dependency contract now includes `verify-extensions`.

### Tests

- **`internal/core/tasks/manifest_write_test.go`** (new) — table-driven coverage for the three helpers, plus an end-to-end case that writes the files and runs the real `LoadValidatedTaskGraphManifest` to prove the mutated manifest is valid.
- **`internal/core/extension/host_writes_test.go`** — `TestHostTasksCreateUpdateIndexAppendsTaskGraphNode` drives the real `host.tasks.create` RPC on a v2 `_tasks.md` and asserts the node + edge injection, the missing `dependencies` key, and that the runner's own validator loads the result.

## Behavior notes

- **Backward compatible.** The legacy Markdown-table `_tasks.md` path is unchanged; only v2 manifests take the new branch.
- **Invalid manifests are never written.** `AppendTaskGraphNode` returns an error if a dependency doesn't resolve to an existing graph node.
- No daemon protocol or host-API surface changes.

## Testing

- **Regression test (red→green):** confirmed `TestHostTasksCreateUpdateIndexAppendsTaskGraphNode` **fails** on the pre-fix code with the exact production error (`tasks table header not found`) and **passes** with the fix.
- `make verify-extensions` — build + `-race` + lint, **0 issues**.
- **Live end-to-end:** with the fix, a real `compozy tasks run` on a v2 workflow reaches `run.started`, injects the QA tasks into the graph, and an agent creates/refactors a file to completion.

## CI note

If the `verify` job fails on a `goconst` in `internal/core/run/ui/multi_remote.go` (`"merged"`), that is a **pre-existing issue on `main`** (main's own CI is currently red on it), not introduced by this change — a rebase clears it once main's lint is green. This branch is rebased on current `main`.

## Summary by CodeRabbit

* **New Features**
  * Added support for creating and updating tasks in v2 task-graph manifests, including correct rendering and graph relationship preservation.
  * Index updates now append nodes and dependency edges while avoiding dependency frontmatter duplication.

* **Bug Fixes**
  * Improved consistency when loading and updating task indexes, including validation that dependencies resolve to existing graph nodes.

* **Chores**
  * Extended the `verify` pipeline to build, test (with race checks), and lint extension modules automatically.
  * Updated ignore rules and added the standalone Go extension module setup.


antoniofulg:
I found some issues when trying locally, I'll put again as ready to review it when I have it working well.

antoniofulg:
Good catch — fixed. Raised `min_compozy_version` to `0.2.14` in `extensions/cy-qa-workflow/extension.toml`.

Caveat: this PR *is* the host-side graph fix, so the "first release containing the new graph branch" isn't tagged yet — the current latest release (v0.2.13) predates it. `0.2.14` is my anticipated next release; please confirm/adjust it to whatever version you actually cut with this change. The `go.mod` require can only reference an existing tag, so it stays on a released version for now — happy to pin it to the fix release once tagged, or to land the host fix first and raise the extension floor + go.mod pin in a follow-up, whichever you prefer.

Without the raise, the newly added `go.mod` lets the extension start on any host ≥ 0.1.10, and its required `plan.pre_discover` → `host.tasks.create` on a v2 `_tasks.md` would still abort with `tasks table header not found` on hosts predating the graph branch — exactly the case you flagged.

antoniofulg:
Re-validated against a real `compozy.tasks/v2` workflow on a canonical (flat) layout: with `cy-qa-workflow` enabled, `host.tasks.create` correctly appends the QA report + execution tasks to the v2 `graph.nodes`/`graph.edges` — no `tasks table header not found` abort and no shadow `_tasks.md` — and the run executes all four jobs (2 implementation + 2 injected QA) to completion. Marking ready for review.

pedronauck (changes requested):
The PR-specific implementation and local gates passed in my audit, but the required GitHub `verify` check on head `e0d78f56654534e2544edea96e7dafb7d26fc7ce` is red, so this cannot be merged under the repository's zero-issue policy. The failing log is `internal/core/run/ui/multi_remote.go:1038:21: string merged has 4 occurrences, make it a constant (goconst)`.

I confirmed that file is identical between this head and current `main`, so this is a base/environment failure rather than a defect introduced by #234. Please update the branch after the `main` Go 1.26.4 lint failure is fixed (or coordinate the base fix first), trigger a fresh CI run, and return with the required `verify` check green. I am not requesting a workaround or unrelated suppression in this PR; the blocking condition is the failed canonical gate.

pedronauck (approved):
Re-review completed at head `1ffb01eeb19285ec9925ea065ff63b7ecee8d79d`. The prior host-version concern is resolved, no active review threads remain, and the current merge ref `87427a84bb861660045df70014905d5be5211993` passed CI run 29539072178 end to end: zero lint issues, 4,775 tests with 4 intentional skips, extension-module verification, build, and Playwright. Approved for squash merge.

pedronauck (approved):
Current head and exact merge tree are clear. Canonical workflow run 29542823245 verified merge SHA 4f4e66a6a908a805707b5e17521aed2240781589: formatting passed, golangci-lint reported 0 issues, 4,798 Go tests passed (4 intentional skips), build and cy-qa-workflow module verification passed, and 5 Playwright E2E tests passed.

PR #243, opened 2026-07-16
fix: add tasks functionality for extensions

renanfferreira:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task table recognition to support varied column ordering, spacing, capitalization, and markdown formatting.
  * Ensured new task entries are added to the correct Tasks table when multiple task-related tables are present.
  * Added clearer handling when the required Tasks table header is missing.

* **Tests**
  * Added coverage for valid and invalid table headers and task row insertion scenarios.

### DIFF
--- a/internal/core/extension/host_writes.go
+++ b/internal/core/extension/host_writes.go
@@ -418,7 +418,7 @@ func appendTaskIndexRow(content string, row string) (string, error) {
 	lines := strings.Split(strings.TrimRight(content, "\n"), "\n")
 	headerIdx := -1
 	for idx, line := range lines {
-		if strings.TrimSpace(line) == "| # | Title | Status | Complexity | Dependencies |" {
+		if isTaskTableHeader(strings.TrimSpace(line)) {
 			headerIdx = idx
 			break
 		}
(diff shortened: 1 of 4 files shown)


pedronauck (changes requested):
Blocking at head `04d638effae8262a0298824fb2dca6628c8379a4`: the new header matcher accepts arbitrary column order, but the production path does not reorder the row. `prepareTaskIndexUpdate` always builds `# | Title | Status | Complexity | Dependencies` via `formatTaskIndexRow`; under the accepted header `# | Title | Complexity | Status | Dependencies`, a real `host.tasks.create` call writes `pending` into Complexity and `high` into Status. The added reordered-header test bypasses this by hand-building an already-reordered row and calling `appendTaskIndexRow` directly, so it cannot catch the production corruption.

Please make the contract consistent: either accept only the canonical exact order documented by this PR, or parse the header and map production row values into that order. Add a regression through `prepareTaskIndexUpdate` or the public host operation using a non-default status and complexity, then resolve the remaining test-suite convention findings and rerun the full `make verify` gate on the final SHA. The current required `verify` check is also red.

pedronauck:
Closing as superseded by #234, which is now merged. #234 adds v2 task creation through the canonical `compozy.tasks/v2` graph (`graph.nodes` and `graph.edges`) and keeps task-owned metadata in task files. This PR would additionally require a mutable Markdown table that duplicates status and dependency state, can drift from the authoritative graph/frontmatter, and its arbitrary-column matcher can write values under the wrong columns. The root task-creation issue is fixed without those duplicate sources of truth.

PR #286, opened 2026-07-31
fix(loop): hydrate externalized node outputs before namespace construction

### LINKED ISSUE #285: Loop fan-out breaks when a node's output exceeds the 16 KB inline limit
## Summary

Once a Loop node's output crosses `LoopOutputInlineLimitBytes` (16 KB), it is externalized to
`loop_output_blobs` and `loop_generation_outputs.output_ref` holds a `sha256:…` ref instead of the
payload. Nothing ever reads that blob back. `runtimeNamespace` builds the namespace from the raw ref,
so every template reaching into a large node's output resolves against the literal string
`"sha256:…"` rather than the value.

The visible failure is a fan-out dying immediately after its producer succeeds:

```
task: coordinator run "run-…" failed: loop: validation failed:
fan-out collection reference ".nodes.load_tasks.output.tasks" is unavailable
```

This is not limited to fan-out. `runtimeNamespace` also feeds gate criteria, action params and prompt
templates, so any `{{ .nodes.<id>.output.<field> }}` silently degrades to a hash string once that
node's output passes 16 KB. Fan-out is simply where it fails loudly.

## Impact

The first-party `dev-cycle` `software-delivery` Loop is broken for realistic inputs. Its `load_tasks`
node calls `ext__dev_cycle__import_tasks`, which inlines each task's full markdown `body`. Four
ordinary task files produced a 41 KB output here, so the fan-out that drives the whole delivery could
not resolve its collection. The run fails ~40 ms in; the coordinator surfaces only
`coordinator_failed` with a generic message, and the real cause appears solely in the daemon log.

## Reproduction

Any Loop whose fan-out collection references a node output larger than 16 KB:

```yaml
- id: produce
  class: action
  kind: <any tool returning > 16 KB>
  produces:
    items: array

- id: spread
  class: control
  kind: fan-out
  collection: "{{ .nodes.produce.output.items }}"
```

I isolated the threshold by running `software-delivery` in a throwaway workspace against a single
synthetic task file of increasing size, reading back `loop_generation_outputs.output_ref`:

| tool output | `output_ref` storage | run status |
| --- | --- | --- |
| 2 713 B | inline | running |
| 9 789 B | inline | running |
| 15 822 B | inline | running |
| 18 142 B | `sha256:…` blob | **failed** |
| 19 302 B | `sha256:…` blob | **failed** |

The boundary matches `LoopOutputInlineLimitBytes = 16 * 1024` exactly.

### Actual behavior

`outputValue` receives a content-addressed ref, fails to parse it as JSON, and returns the raw
`"sha256:…"` string. `valueAtPath` cannot walk `.output.tasks` inside a string, returns `!ok`, and
`resolveFanOutCollection` raises `fan-out collection reference … is unavailable`. The run terminates
with `coordinator_failed`.

### Expected behavior

A node output larger than the inline limit resolves exactly like a smaller one. Externalization is a
storage detail and should be invisible to templates, fan-out collections and gate criteria.

## Root cause

`internal/loop/control_namespace.go` builds each node entry straight from the ref:

```go
entry["output"] = outputValue(output.OutputRef)
```

and `outputValue` has no notion of a content-addressed ref:

```go
func outputValue(ref string) any {
	trimmed := strings.TrimSpace(ref)
	if trimmed == "" {
		return nil
	}
	var value any
	if err := decodeSingleJSONValue(strings.NewReader(trimmed), &value); err == nil {
		return value
	}
	return trimmed // "sha256:…" falls through as a plain string
}
```

The read side already exists — `getLoopOutputByRefWithExecutor` in
`internal/store/globaldb/global_db_loop_outputs.go`, plus `ErrOutputRefNotFound` in the loop package.
It is simply never called on this path: `writeGenerationOutputBlob` persists the payload and nothing
hydrates it back.

Note that `output_ref` already holds *either* inline JSON *or* a ref depending on size, so resolving
a ref into its payload before namespace construction restores the invariant `outputValue` has always
assumed rather than introducing new semantics.

## Acceptance criteria

- Node outputs above `LoopOutputInlineLimitBytes` resolve in templates identically to inline outputs.
- Fan-out collections referencing an externalized output materialize their branches.
- Fan-out materializations that themselves exceed the limit are resolved too, so `fanOutItem` keeps
  serving `.item` to the fan-out body.
- Gate criteria, action params and prompt templates reading `{{ .nodes.<id>.output.<field> }}` behave
  the same above and below the limit.
- A missing blob surfaces `ErrOutputRefNotFound` rather than degrading to a hash string.
- The HTTP surface (`GET /loop-runs/:id`) keeps returning refs, not hydrated payloads.
- Regression coverage for a node output on both sides of the limit.

## Scope

Read-path only. The write path (`OutputPayloadRequiresRef`, `writeGenerationOutputBlob`) and the
16 KB limit itself are unchanged — the limit is a reasonable inline cap, it just needs a matching
read. Hydration belongs in the `loop` package next to where outputs are loaded, not in the store
repository, so the API surface keeps returning refs.

## Environment

- Compozy `0.3.0-beta.2`, macOS 15 (arm64), SQLite store
- Loop: `dev-cycle` `software-delivery` (extension `dev-cycle` 0.3.1)
- Present on `main` as of `599b32d` — none of the 15 commits between `v0.3.0-beta.2` and `main` touch
  this path, so the v0.3.0 release PR (#272) does not fix it.

## Proposed fix

Draft PR: https://github.com/compozy/compozy/pull/286

magoamz:
Fixes #285.

## Problem

Loop node results larger than `LoopOutputInlineLimitBytes` (16 KB) are externalized to
`loop_output_blobs`, and `GenerationOutput.OutputRef` then holds a `sha256:…` ref rather than the
payload. Nothing read that blob back. `runtimeNamespace` builds each node entry straight from the
ref:

```go
entry["output"] = outputValue(output.OutputRef)
```

`outputValue` tries to parse the ref as JSON, fails, and returns the raw `"sha256:…"` string. Every
template reaching into that node's output then resolves against a hash.

Fan-out is where it fails loudly — `resolveFanOutCollection` cannot walk `.output.tasks` inside a
string, so the run dies with `fan-out collection reference … is unavailable` right after its producer
succeeded. Gate criteria, action params and prompt templates degrade silently through the same path.

This breaks the first-party `dev-cycle` `software-delivery` Loop for realistic inputs: `load_tasks`
inlines each task's full markdown body, so four ordinary task files produced a 41 KB output and the
fan-out driving the whole delivery could not resolve.

## Change

Resolve content-addressed refs into their payloads immediately after the coordinator loads generation
outputs, so externalization stays a storage detail instead of something every consumer must know
about.

The read side already existed and was simply never wired — `getLoopOutputByRefWithExecutor` had no
production caller, and `ErrOutputRefNotFound` was already defined for exactly this. This exposes it
as `LoopRepo.GetGenerationOutputPayload`, adds it to `GenerationOutputReader`, and calls it from
`hydrateGenerationOutputs`.

Two properties worth calling out:

- **Fan-out materializations are covered too.** A materialization carrying large items exceeds the
  limit itself and becomes a blob, which would break `fanOutItem` serving `.item` to the fan-out
  body. Hydrating at the load site fixes the collection and `.item` in one place; fixing only
  `resolveFanOutCollection` would have failed at the next step.
- **The HTTP surface is unchanged.** Hydration is scoped to the two coordinator load sites rather
  than to `LoopRepo.ListGenerationOutputs`, so `GET /loop-runs/:id` keeps returning refs instead of
  inlining full payloads into API responses.

`output_ref` already holds *either* inline JSON *or* a ref depending on size, so replacing a ref with
its payload restores the invariant `outputValue` has always assumed rather than introducing new
semantics.

## Scope

Read-path only. The write path (`OutputPayloadRequiresRef`, `writeGenerationOutputBlob`) and the
16 KB limit itself are untouched — the limit is a reasonable inline cap that was just missing its
matching read.

## Tests

`internal/loop/generation_output_hydration_test.go` covers:

- a content-addressed ref replaced by its payload
- inline outputs passing through with no blob lookup
- one lookup per distinct ref when several outputs share a payload
- a missing blob surfacing `ErrOutputRefNotFound` instead of degrading to the ref
- a regression asserting the fan-out collection reference is unavailable before hydration and
  resolves to its items after

```
go test ./internal/loop/...
ok  github.com/compozy/compozy/internal/loop
ok  github.com/compozy/compozy/internal/loop/dsl
ok  github.com/compozy/compozy/internal/loop/dsl/refs
ok  github.com/compozy/compozy/internal/loop/gate
ok  github.com/compozy/compozy/internal/loop/goal
ok  github.com/compozy/compozy/internal/loop/watch
```

`go vet ./internal/loop/... ./internal/store/globaldb/... ./internal/daemon/...` is clean, as is
`gofmt`.

## Verification against the reported failure

The 16 KB boundary was measured by running `software-delivery` in a throwaway workspace against a
synthetic task file of increasing size and reading back `loop_generation_outputs.output_ref`: outputs
of 2 713 B / 9 789 B / 15 822 B stayed inline and ran, while 18 142 B / 19 302 B became blobs and
failed. Details in #285.

I have not been able to run the full repository suite locally (the sandbox only has the Go packages
checked out), so CI is the real signal for anything outside `internal/loop`.

### DIFF
--- a/internal/loop/coordinator.go
+++ b/internal/loop/coordinator.go
@@ -2,6 +2,7 @@ package loop
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"strings"
@@ -45,6 +46,9 @@ type GenerationOutputReader interface {
 		runID RunID,
 		generation int,
 	) ([]GenerationOutput, error)
+	// GetGenerationOutputPayload resolves one content-addressed output ref into the payload
+	// the producing node returned.
+	GetGenerationOutputPayload(ctx context.Context, outputRef string) (json.RawMessage, error)
 }
 
 // GateDecisionReader reads persisted human decisions for coordinator gate re-evaluation.
@@ -196,6 +200,9 @@ func (r *CoordinatorRunner) buildCoordinatorPlan(
 	if run.Generation > 0 {
 		outputs, err := r.outputs.ListGenerationOutputs(ctx, run.WorkspaceID, run.ID, run.Generation)
 		if err != nil {
+			return task.CoordinatorCompletionPlan{}, err
+		}
+		if err := hydrateGenerationOutputs(ctx, r.outputs, outputs); err != nil {
 			return task.CoordinatorCompletionPlan{}, err
 		}
 		if len(outputs) > 0 {
--- a/internal/loop/coordinator_action.go
+++ b/internal/loop/coordinator_action.go
@@ -67,6 +67,9 @@ func (r *CoordinatorRunner) ExecuteActionRun(
 	if err != nil {
 		return task.RunResult{}, err
 	}
+	if err := hydrateGenerationOutputs(ctx, r.outputs, outputs); err != nil {
+		return task.RunResult{}, err
+	}
 	input, err := actionExecutionInput(
 		taskRun,
 		actor,
--- a/internal/loop/coordinator_test.go
+++ b/internal/loop/coordinator_test.go
@@ -2,6 +2,7 @@ package loop
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"io"
 	"log/slog"
@@ -2433,7 +2434,8 @@ func (r *coordinatorRunnerTaskRunReader) GetTaskRun(
 }
 
 type coordinatorRunnerOutputs struct {
-	outputs map[int][]GenerationOutput
+	outputs  map[int][]GenerationOutput
+	payloads map[string]json.RawMessage
 }
 
 func (r coordinatorRunnerOutputs) ListGenerationOutputs(
@@ -2446,6 +2448,17 @@ func (r coordinatorRunnerOutputs) ListGenerationOutputs(
 		return nil, nil
 	}
 	return append([]GenerationOutput(nil), r.outputs[generation]...), nil
+}
+
+func (r coordinatorRunnerOutputs) GetGenerationOutputPayload(
+	_ context.Context,
+	outputRef string,
+) (json.RawMessage, error) {
+	payload, ok := r.payloads[outputRef]
+	if !ok {
+		return nil, ErrOutputRefNotFound
+	}
+	return payload, nil
 }
 
 type coordinatorRunnerLoopStore struct {
--- a/internal/loop/generation_output_hydration.go
+++ b/internal/loop/generation_output_hydration.go
@@ -1,0 +1,47 @@
+package loop
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+// hydrateGenerationOutputs replaces content-addressed output refs with the payloads they
+// stand for, in place.
+//
+// Outputs at or below LoopOutputInlineLimitBytes are persisted inline, so OutputRef already
+// carries the value; larger ones are externalized to the output blob store and OutputRef
+// carries a "sha256:…" ref instead. Everything downstream — runtimeNamespace, fan-out
+// materialization, gate criteria, action params — reads OutputRef as the value, so an
+// unresolved ref silently degrades a node's output to a hash string. Resolving refs here
+// keeps externalization a storage detail rather than something every consumer must know
+// about.
+func hydrateGenerationOutputs(
+	ctx context.Context,
+	reader GenerationOutputReader,
+	outputs []GenerationOutput,
+) error {
+	if reader == nil {
+		return nil
+	}
+	// One blob can back several outputs (a fan-out branch reusing its producer's payload,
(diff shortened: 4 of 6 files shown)


pedronauck:
Closing this PR because its externalized Loop-output hydration fix was incorporated and extended in #305, which has now been merged as the complete Loop node lifecycle landing change. The merged implementation preserves inline outputs, hydrates external references before namespace evaluation, caches loaded payloads, covers fan-out and gate consumers, and reports missing blobs deterministically. @magoamz is credited as a co-author on the final merged commit. Thank you for the original fix and investigation.

PR #352, opened 2026-08-12
fix: support authenticated LAN browser pairing

cotts:
## Summary

- Enable authenticated browser pairing over a trusted local network without Tailscale or external internet traffic.
- Keep the LAN HTTP listener protected by source allowlisting and device authentication after the one-time pairing bootstrap.
- Document the LAN/DMZ deployment path alongside the existing Tailscale gateway flow.

### LAN browser pairing

Remote HTTP requests previously applied device authentication before the browser could redeem its first pairing artifact. The pairing redemption route is now registered before that middleware for the LAN HTTP surface; all other remote routes remain protected.

### Workspace configuration loading

Global configuration is no longer loaded twice when the workspace overlay resolves to the same file, avoiding the `gateway settings are global-only` startup failure.

### Documentation and QA

The Gateway quickstart, security guide, configuration references, official Compozy skill, and pairing scenario now describe trusted LAN access, explicit CIDR restrictions, one-time pairing redemption, and deployment in an isolated network or DMZ.

## Verification

- `make gate` — PASS; full gate record is current for the committed tree
- 21,490 Go tests passed; 3 expected platform skips
- Bun formatting, lint, typecheck, tests, and web build passed
- Go lint, build, and boundaries passed
- Direct LAN verification returned `HTTP 200` for the UI and reached the pairing endpoint without a device session
- Tailscale provider was disabled on the deployment host; no external relay is required for the LAN path

## Compozy Impact Audit

- Native tools: no native tool IDs, toolsets, descriptors, schemas, digests, risk flags, or capability gates changed; the existing gateway management surfaces remain the source of truth.
- Extensibility and hooks: no extension, hook, bridge SDK, MCP sidecar, registry, or provider contract changed. The Tailscale provider is optional; LAN access uses the existing HTTP configuration lifecycle and pairing flow.
- Workspace data isolation: the changed HTTP middleware is global/operator-scoped; pairing credentials remain device-scoped, and no workspace-scoped data is exposed before authentication. The workspace config loader now avoids reapplying the same global file.
- Official Compozy skill: updated `skills/compozy/references/configuration.md` and `skills/compozy/references/runtime-operations.md` with the LAN bootstrap and authentication behavior.

## Test plan

- [x] `make gate`
- [x] LAN UI request and unauthenticated pairing-endpoint reachability check
- [x] Regression test: LAN pairing redemption bypasses only device bootstrap authentication
- [x] Server deployment and binary hash verification

## Summary by CodeRabbit

* **New Features**
  * Added settings to enable remote HTTP access and restrict it to specific IP addresses or networks.
  * Added a General Settings interface for configuring remote access and allowed IPs.
  * Added validation for IP addresses and CIDR ranges.
  * Remote pairing remains available while protected actions require authentication.

* **Bug Fixes**
  * Corrected cookie security behavior for non-TLS connections.
  * Prevented duplicate configuration loading when global and workspace paths match.


pedronauck (changes requested):
Thanks @cotts — this PR lands on a real need, and it's impressively aligned with the repo's process for an external contribution: the Impact Audit is filled in, docs and the official skill are co-shipped, CLI config keys are wired, the settings UI/API/OpenAPI are regenerated, the QA scenario is touched, and `make gate` evidence is included. The demand is also legitimate by our own design record: the archived gateway PRD explicitly anticipated this deployment — *"A direct network bind onto an operator-managed network is subject to the same rules: authentication active, pairing gate served, no exceptions"* (`.compozy/tasks/_archived/remote-gateway/_prd.md:81`). We want this feature.

I'm requesting changes because the implementation sits on the wrong chassis: it exposes the **local** surface set on the network instead of modeling LAN as a remote gateway surface, and several of the changes regress or bypass load-bearing security invariants. After reviewing the alternatives, the direction we want is the one our own architecture already prescribes: **LAN access ships as a bundled connectivity provider extension, not as core HTTP changes.** ADR-003 in the gateway design record states it directly: *"Exposure policy lives in core; connectivity mechanisms ship as provider extensions."* A LAN bind is a connectivity mechanism. Full shape below — and the good news is it deletes most of this diff's core surface instead of asking you to harden it, while your route-ordering instinct (redeem before device auth) is already how the private tier works (`internal/api/httpapi/surface_routes.go:27-37`).

---

## Why the current shape can't merge

### 1. The LAN path exposes `SurfaceSetLocal`, which is strictly more than any remote tier is allowed to serve

The gateway's security model is *"a tier is a property of the listener, not of the request"* — each listener registers a fixed route matrix (`internal/api/httpapi/surface_routes.go:9-25`). The private (Tailscale) tier serves: redeem → device auth → operator routes **without** local task lifecycle (`registerOperatorRoutes(api, handlers, false)`), **read-only** resources (`registerRemoteResourceReadRoutes`), gateway management, stream tickets. It deliberately excludes:

- `registerAgentKernelRoutes` (`internal/api/httpapi/agent_routes.go:5-24`): `POST /api/agent/spawn`, `POST /api/agent/tasks/claim-next`, claim-token lifecycle, Network channel send/recv. These are process-trust primitives whose caller identity is inferred from the calling process environment (`internal/agentidentity`) — there is no meaningful browser identity for them, and `claim-next` is the authoritative claim primitive. Exposing them to any network peer (even a device-authenticated one) is a capability escalation, not a feature. No browser needs these routes.
- Task lifecycle mutations (`registerTaskRoutes(..., true)` → start/cancel/approve/reject/scheduler drain, `internal/api/httpapi/routes.go:208-252`).
- Full resource CRUD (`PUT`/`DELETE /api/resources/:kind/:id`, `routes.go:107-123`).
- Webhook ingress + `GET /api/mcp/oauth/callback`, which in this PR are registered **before** the remote guard (`routes.go:17-18`) and therefore sit on the LAN with no IP allowlist and no device auth.

The PR's LAN branch keeps all of the above (the new `switch` in `routes.go` registers redeem + `remoteDeviceAuthMiddleware` and then falls through to the unchanged local registrations). Net effect: **a LAN-paired browser gets a strict superset of what a Tailscale-paired device gets — over plaintext HTTP.** It also silently falsifies the published invariant in `packages/site/content/docs/gateway/security.mdx` ("The agent kernel, hosted MCP, task mutations, and resource mutations exist only on local access and never move"), which this PR does not update. On top of that, `loopbackMutationGuardWithRemote(boundHost, allowRemote)` opens the privileged mutation set (settings PATCH/reload/restart, vault secrets, extension installs, tool invoke, bridges) to remote devices — including the ability to rewrite `http.allowed_ips`/`allow_remote_access` themselves, making the security boundary self-widening from outside it.

### 2. `Secure: c.Request.TLS != nil` regresses the existing Tailscale path

On `main`, `Secure: true` (`internal/api/httpapi/gateway_redeem.go:33`) is correct *by construction of the listener*, not by inspecting the request: every tier listener binds `127.0.0.1` (`internal/daemon/gateway_tier_server.go:38`) and TLS terminates inside the embedded tsnet node (`extensions/connectivity/tailscale/node.go:71-81`), which forwards plain HTTP to loopback. That means `c.Request.TLS` is **nil on the Tailscale tiers too** — so this change strips the `Secure` attribute from cookies issued on the existing HTTPS private/public paths, not just on the LAN. Your own test flips confirm it: `TestGatewayTierAuthentication` now asserts `Secure == false` for tier redemption (`gateway_routes_test.go`), and `skills/compozy/references/runtime-operations.md` still promises a `Secure` cookie two lines below the new LAN paragraph.

The underlying problem you were solving is real — with `Secure: true`, a browser on `http://192.168.x.y` silently drops the cookie and every subsequent request 401s. Under the provider shape below, the provider terminates TLS on the LAN just like tsnet does on the tailnet, so `Secure: true` stays untouched and the problem never exists.

### 3. The exposure switch lives in workspace-overridable config

`[http].*` is not protected by the workspace overlay validation — `validateWorkspaceConfigOverlay` (`internal/config/workspace_overlay.go:41-52`) rejects only `gateway` and `marketplace` tables, and `ValidateConfigWriteScope` (`internal/config/write_scope_policy.go:17-23`) likewise keys on the first path segment. As written, a checked-out repository's workspace config file can set `http.allow_remote_access = true` + `http.host = "0.0.0.0"` and the daemon exposes itself on the next restart. That contradicts the PRD's hard rule ("no flag, environment value, or configuration combination bypasses this", `_prd.md:79-81`) and `internal/config/gateway_config.go:27-28` ("Durable exposure intent lives in the global database, never in this config").

Under the provider shape this problem dissolves entirely: exposure intent lives where it already does — the durable gateway state (provider activation + surface exposure, `internal/store/globaldb/schema/definitions/61_gateway.sql:23-47`), written through generation-fenced transitions, never through config at all.

---

## Target shape: a bundled LAN connectivity provider extension

A small bundled provider at `extensions/connectivity/lan/` — a listener on the configured LAN interface forwarding to the loopback private-tier listener, exactly like the Tailscale provider forwards tailnet traffic. The Tailscale extension is the working reference implementation: out-of-process, enrolled at boot (`internal/daemon/boot_automation_extensions.go:31`), implementing the `ConnectivitySource` contract (`internal/gateway/connectivity.go:89-99`), with the loopback forward target already enforced on the provider side (`extensions/connectivity/tailscale/provider_endpoint.go:33`).

Everything hard about this feature is then already built and stays untouched:

- **Route surface**: the private-tier matrix (pairing gate → device auth → operator surface, read-only resources, stream tickets) is exactly what a LAN browser should get. Zero route changes. The `"tier listeners must bind a loopback host"` invariant (`internal/api/httpapi/server_setup.go:82-83`) stays intact.
- **Lifecycle + consent**: the reconciler's desired/observed state machine, generation fencing, and the provider enable flow (install-source trust + control digest re-checked on every enable and boot) are the enable UX — `compozy gateway providers enable lan --tier private` + `compozy gateway surface enable`, same as Tailscale. Agent-manageable through the existing `/api/gateway/*` surface for free.
- **Audit + status**: an advertised LAN endpoint flips `auditLocalOnly` (`internal/gateway/audit.go:120-127`) and appears in `Status.Addresses` through the existing provider-endpoint projection. `compozy gateway audit` sees the exposure with no new plumbing.
- **Transport security**: the provider terminates TLS itself with a daemon-generated local CA (the LAN analogue of tsnet's tailnet certificates), and the certificate fingerprint travels alongside the pairing artifact — which already moves through a trusted channel. This makes greptile's plaintext-credential finding (which I'm confirming as valid against the current diff) **disappear** instead of being consented: the reusable device cookie never crosses the LAN in cleartext, and `Secure: true` needs no change.
- **Cookies, pairing, streams**: `Secure`/`HttpOnly`/`SameSite` untouched; pairing source stays `private`; the web client is already tier-aware and mints stream tickets per connect (`web/src/lib/gateway-stream-auth.ts`).
- **Remote administration posture**: inherits exact parity with the Tailscale tier — no new decisions, no new flags. (Separately from this PR, we'll open a follow-up issue on making `privilegedMutationGuard` surface-aware and adding a boundary-key deny-list for remote origins; that applies to both transports and is out of scope here.)

### Provider implementation notes

1. **Source allowlist lives in the provider.** CIDR allowlist enforced at the provider's own listener. Fail-closed default: when unset, allow only private ranges (RFC1918 + link-local + loopback) rather than any reachable source — that matches the "my internal network" intent with zero configuration. The private-prefix table already exists twice in the tree (`internal/outboundpolicy/policy.go:228-254`, `internal/bridges/contract/webhook.go:36-50`); extract a shared exported helper rather than adding a third copy. And call `addr.Unmap()` before matching — `netip.Prefix.Contains` is address-family strict, so a dual-stack accept handing you `::ffff:192.168.1.42` will not match `192.168.1.0/24` otherwise.
2. **Sanitize forwarded headers.** The daemon trusts `X-Forwarded-Proto` from its forwarder (`internal/api/httpapi/middleware.go:214-230`), which is safe only because the forwarder is Compozy-owned. The provider must strip client-supplied `X-Forwarded-*`/`Forwarded` headers and set its own values before proxying, or a LAN client can steer scheme classification into the CORS/target checks.
3. **One provider per tier is a known, acceptable constraint.** `gateway_active_provider_per_tier` (`61_gateway.sql:34-36`) means LAN and Tailscale are mutually exclusive on the private tier. That fits this PR's stated deployment (isolated network / DMZ, Tailscale disabled); hybrid setups are a future relaxation if demand shows, not this PR's problem.
4. **The only core touch this needs**: endpoint verification. The challenge probe is forced to HTTPS (`internal/gateway/endpoint_verify.go:260-262`) and would refuse a locally-issued certificate, so the `ConnectivitySource` contract needs a provider-supplied trust anchor (or pinned fingerprint) for the probe — a small, well-scoped contract accommodation with `EndpointSchemePolicyTailnetInternal` (`internal/gateway/connectivity.go:32-33`) as the precedent for tier-scoped verification policy. Nothing else in core should change.

### Why not core HTTP changes (the shape this PR chose)

We considered a core variant (a dedicated LAN listener with its own surface set). It would require relaxing the loopback invariant, a parallel admission path outside the reconciler, and new audit/status/parity plumbing — a permanent enlargement of the core security surface to duplicate what the provider machinery already does. The extension shape keeps core's invariants byte-for-byte and validates the provider contract with a second real implementation, which is exactly why that contract exists. The less this feature touches core, the better.

---

## What this means for the current diff

- **Revert the core/HTTP surface**: `[http].allow_remote_access` + `allowed_ips` (config, overlay, defaults, CLI keys, contracts, OpenAPI, generated TS), the `routes.go`/`middleware.go`/`gateway_auth.go` remote-guard changes, the `gateway_redeem.go` Secure change and its test flips, the `daemon/settings.go` parity change, and the web settings HTTP section (including the `null → []` normalization shim — with the config keys gone, nothing needs it). The extension shape doesn't need any of it.
- **Keep and adapt the docs work**: the Gateway index/quickstart/security pages and the official skill references you updated are genuinely useful — rewritten for the provider flow (`providers enable lan` + pairing) instead of `[http]` keys. The security table gains an honest row ("Private overlay, carried by the LAN provider — devices on your local network") without breaking the cross-table invariants sentence.
- **Keep the QA scenario touch**: `docs/qa/scenarios/RT-gateway-paired-device.md` still needs its LAN leg — rewritten for the provider flow, and per the repo's QA contract the changed scenario needs a recorded walk (or an explicit `blocked-verify` rationale) before this closes.
- **Split out the config-load fix now**: the duplicate global/workspace overlay load fix (`internal/config/config_load.go:123-130`) is real, independently correct, and unblocks your deployment on its own (it's what caused the "gateway settings are global-only" boot failure). Open it as its own small PR and it can merge quickly — with the raw string comparison upgraded to canonicalized paths (`filepath.EvalSymlinks`-style) so a symlinked config is recognized too, plus the repo's `t.Run("Should …")` test pattern.

---

## Summary

- **Direction**: LAN access ships as a bundled connectivity provider extension (`extensions/connectivity/lan/`), reusing the private tier end-to-end — provider-terminated TLS with a local CA, CIDR allowlist in the provider (fail-closed to private ranges), forwarded-header sanitization, and one small core contract accommodation for challenge-probe trust. No core route, invariant, cookie, or config-surface changes.
- **From this diff**: revert the core/HTTP/web changes; adapt the docs + skill + QA scenario to the provider flow; split the config-load fix into its own PR (mergeable now).
- **Why**: the current shape exposes the local surface (agent kernel, task mutations, resource CRUD, privileged mutations) to the network over plaintext, regresses the Tailscale cookie, and puts the exposure switch in workspace-overridable config — while the provider shape gets pairing-gated LAN access with the same guarantees Tailscale users have today, and mostly by deleting code from this diff rather than adding more.

Happy to pair on the provider — the Tailscale extension is a close template, and the touch list above maps the one contract accommodation needed in core. Your deployment need is exactly the use case this should serve, and the config-load fix PR would land immediate relief while the provider comes together.

🤖 Review prepared with [Claude Code](https://claude.com/claude-code)

cotts:
Instead of an internal Compozy feature, it can be used with a reverse-proxy without break any project security rule.

PR #396, opened 2026-08-14
fix: guard autonomous memory extractor writes

### LINKED ISSUE #395: Memory extractor persists forbidden operational chatter and overwrites unrelated entries on generated slug collisions
## CompozyOS version

`v0.3.0-beta.13`

The same relevant controller, scanner, and extractor code is still present on `main` at `598ed807c1b4eec70618cf4e3476236b5796fed4`.

## Operating system

macOS (Apple Silicon)

## Area

- Core / daemon
- Sessions
- Workspaces

## ACP agent

Claude Code

## Model / provider

Claude Sonnet 5

## Steps to reproduce

1. Start a session with Memory v2 transcript extraction enabled in a workspace that already contains a project-scoped memory, for example `project_example.md`.
2. Discuss a rejected `memory_propose` operation in the session, including controller details such as `policy_repo_path` or `memory.write.rejected`.
3. Allow the background extractor to process the session.
4. Inspect the workspace memory and memory admin history.
5. Observe that the extractor can persist operational metacontext about its own controller, despite `extract.v1.tmpl` instructing it to apply `WHAT_NOT_TO_SAVE`.
6. Have the extractor produce a different fact with the same broad entity used by an existing memory (for example `example`).
7. Observe that the generated filename collides with `project_example.md` and the existing entry can be updated even when the new content is semantically unrelated.

## What happened

The extractor persisted an entry equivalent to:

> Extractor rejected a project memory update via memory_propose with reason policy_repo_path; the write was blocked by a scan rule.

This is transient operational chatter about Memory v2 itself, not durable project knowledge.

In the same session, an existing project memory was replaced by a different fact after the extractor generated the same target slug from a broad entity. The manual write being discussed was rejected and was not responsible for the replacement; the memory event chronology showed that the extractor update happened earlier.

The scanner does run for extractor candidates, but its deterministic rules do not cover this class of Memory-controller/extractor metacontext. For example, `policy_repo_path` only matches selected directory prefixes, and the operational-state patterns do not match terms such as `memory_propose`, `memory.write.rejected`, or `policy_repo_path`.

There is also a separate matching hazard in `internal/memory/controller/controller.go`: `exactFilenameTarget()` is handled before semantic/entity ambiguity checks. For an autonomous write whose filename was generated from model-supplied `type + entity`, an exact slug collision can therefore become an update without proving that the candidate belongs to the same durable knowledge slot.

`memory.write.shadowed` events were not treated as persisted files in this report: the source maps controller `noop` decisions to that event name.

## Expected behavior

- Transcript extraction should not persist operational chatter about the extractor, controller, scan decisions, queues, or memory tools.
- Autonomous candidates should not overwrite an existing entry merely because a model-generated slug collides with its filename.
- Explicit user/CLI updates that intentionally target a filename should retain their current update behavior.

## Suggested regression coverage

1. Scanner/controller test: an extractor candidate describing `memory_propose` rejection and `policy_repo_path` is rejected as `WHAT_NOT_TO_SAVE`.
2. Controller test: an autonomous candidate with a generated filename that collides with an existing entry, but does not explicitly target that entry, must not update it.
3. Controller test: an explicit CLI/user update targeting the same filename still updates it.

## Additional context

This appears to be two interacting safety gaps:

1. The prompt contains the correct semantic policy, but the deterministic scanner only implements a limited set of textual patterns, so model noncompliance can pass through.
2. Generated filename equality is currently treated as sufficient update identity for autonomous writes.

No secrets, local paths, workspace identifiers, session identifiers, or proprietary memory contents are included in this report.

davioliveeira:
## Summary

- reject Memory v2 operational identifiers such as `memory_propose`, native `compozy__memory_*` tool names, controller event names, and scanner rule IDs
- prevent autonomous extractor/provider/dreaming candidates from updating an existing memory solely because their generated slug collides
- preserve explicit filename collision updates for direct CLI/user writes

## Root cause

The extractor prompt described the correct `WHAT_NOT_TO_SAVE` policy, but the deterministic scanner did not recognize operational Memory v2 identifiers. Model-generated operational chatter could therefore reach curated memory.

Separately, exact filename collisions were treated as update identity before semantic matching. For autonomous candidates, the filename is generated from model-supplied type/entity data, so unrelated facts could overwrite an existing entry with the same slug.

Autonomous generated collisions now produce a deterministic `noop` and do not invoke the LLM tiebreaker. This also prevents a tiebreaker `add` result from writing over the colliding filename. Explicitly targeted writes keep the existing update behavior.

Fixes #395.

## Validation

- `go test ./internal/memory/scan ./internal/memory/controller`
- `go test ./internal/memory/...`
- `go test -race ./internal/memory/scan ./internal/memory/controller`
- independent diff review: no remaining actionable findings

`make gate-full` was also attempted twice. Its Go/module verification completed, but `CodegenCheck` is blocked in this checkout by the local Bun runtime (`1.1.22-canary`) differing from the repository's declared `bun@1.3.4`; OpenAPI generation fails while resolving TypeScript/Redocly dependencies. `bun install --frozen-lockfile` was run and the gate retried once, but the environment-level failure remained. No generated or dependency files are included in this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented autonomous memory updates from overwriting existing entries when generated filenames collide.
  * Preserved existing entries without creating duplicate updates in collision scenarios.
  * Added safeguards to reject internal memory-operation identifiers from being saved as regular memory content.
  * Improved collision handling, including numeric-prefix filenames and cases where alternative actions might otherwise be suggested.


pedronauck (changes requested):
Changes requested

1. Cover the full canonical Memory v2 operational surface in the scanner. The new closed matcher in internal/memory/scan/scan.go:238 does not match canonical identifiers such as compozy__memory_note, compozy__memory_search, compozy__memory_reindex, memory.write.reverted, or policy_memory_operational_state itself. Those candidates reach Controller.Decide and can be persisted, which contradicts this PR goal. Derive the matcher from the canonical inventory or cover the controlled family, and add isolated table-driven regression cases for every identifier family.

2. Complete the collision-boundary regression coverage. The new controller tests exercise only OriginExtractor with a generated filename. Cover Provider and Dreaming as autonomous origins, and prove that an explicitly supplied target filename still permits the intended direct update. This protects both sides of the new guard.

The reported lack of t.Parallel is not valid; the new subtests already run in parallel.

PR #400, opened 2026-08-14
fix: bound session prompt payloads

### LINKED ISSUE #399: Web UI resends the full session transcript until prompt requests exceed 4 MiB
## CompozyOS version

`v0.3.0-beta.13` (also present on current `main`)

## Operating system

macOS (Apple Silicon)

## Area

Web UI, Sessions, Core / daemon

## ACP agent

Claude Code

## Steps to reproduce

1. Start or resume a session with a sufficiently large transcript (tool output makes this easier to reach).
2. In the Web UI composer, send any short follow-up message such as `hi`.
3. Observe the inline error `request body too large`.
4. Check the daemon log and observe `POST /api/workspaces/:workspace_id/sessions/:session_id/prompt` returning HTTP 413.
5. Send the same follow-up with `compozy session prompt <session-id> "hi"`; it succeeds because the CLI sends only the new prompt input.

## What happened, and what did you expect?

The Web UI becomes unable to send any further prompt once the serialized transcript exceeds the daemon's 4 MiB API request-body limit. Even a two-character follow-up is rejected before it reaches the ACP provider.

The Web UI should submit only the newly authored user message, matching the session prompt endpoint and CLI behavior. Existing session history is already durable and provider-owned; it should not be uploaded again with every turn.

## Root cause

`web/src/systems/session/lib/session-prompt-chat-transport.ts` correctly finds `latestUserMessage(messages)`, but then builds the request with the full `messages` array:

```ts
const body = {
  idempotency_key: idempotencyKey,
  message_id: messageId,
  messages,
  // ...
};
```

The daemon applies `maxAPIRequestBodyBytes = 4 << 20` to all `/api/*` requests in `internal/api/httpapi/middleware.go`, so the request is rejected with 413 before prompt dispatch. The prompt contract already supports a single `message`, and `ExtractPromptInput` only accepts one user message as the authored input.

In the reproduced session, `compozy session history ... -o json` was about 3.3 MB; the richer UI message serialization pushed the prompt request past 4 MiB.

## Suggested regression coverage

- Build a prompt request from a large transcript plus one small new user message.
- Assert that the request body contains only that newest user message.
- Assert that the serialized request stays bounded by the new input size rather than total transcript size.
- Preserve the newest message ID as `message_id` and idempotency-key reuse on retry.

## Final checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release.

davioliveeira:
## Summary

- send only the newest authored user message from the Web session transport
- keep the durable `message_id`, runtime snapshot, and retry idempotency behavior unchanged
- add a regression test with a historical transcript larger than the daemon's 4 MiB API body limit

## Root cause

The session chat transport selected the latest user message for identity, but serialized the entire `messages` transcript into every prompt request. Once the transcript crossed the daemon's fixed API request-body limit, even a tiny follow-up was rejected with HTTP 413 before reaching the ACP provider.

Session history is already durable and the prompt endpoint extracts one newly authored user message, so resending historical UI messages is unnecessary.

## Validation

- `bun run --cwd web test:raw src/systems/session/lib/__tests__/session-prompt-chat-transport.test.ts` — 5 passed
- `bun run --cwd web typecheck:raw` — passed
- `bunx oxfmt --check web/src/systems/session/lib/session-prompt-chat-transport.ts web/src/systems/session/lib/__tests__/session-prompt-chat-transport.test.ts` — passed
- `bunx oxlint --deny-warnings web/src/systems/session/lib/session-prompt-chat-transport.ts web/src/systems/session/lib/__tests__/session-prompt-chat-transport.test.ts` — passed
- `git diff --check` — passed

The full Web test run was also attempted. It reported two unrelated existing failures in `e2e/fixtures/__tests__/runtime.test.ts` and then did not terminate, so it was interrupted. No files involved in those failures are changed here.

Fixes #399

### DIFF
--- a/web/src/systems/session/lib/__tests__/session-prompt-chat-transport.test.ts
+++ b/web/src/systems/session/lib/__tests__/session-prompt-chat-transport.test.ts
@@ -8,6 +8,12 @@ const userMessage = (id: string, text = "Continue the durable transcript") => ({
   role: "user" as const,
 });
 
+const assistantMessage = (id: string, text: string) => ({
+  id,
+  parts: [{ text, type: "text" as const }],
+  role: "assistant" as const,
+});
+
 function streamResponse() {
   return new Response(
     [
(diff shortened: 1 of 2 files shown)


pedronauck:
since there's no update here, I'll close this PR for now

PR #467, opened 2026-08-25
fix: route command palette entities directly

pedronauck:
## Summary

Linear ENG-131: `[Web] Command palette actions should go direct`.

Concrete rows in the OS command palette now open the entity's identity-bearing detail/action target in one selection. Catalog commands remain catalog commands, while root search and pushed domain views share the same landing behavior.

## Root cause

Most domain row seeds used `searchRoute` or `tasksRoute`, which produced a catalog pathname with a `q`/`query` filter. The row therefore discarded the entity identity and made the operator select the same entity again. The two palette openers also had separate behavior, and foreign-workspace rows could be opened before the window-manager workspace fence was ready.

## Implementation

- Replaced catalog-plus-filter entity routes with explicit routes for Tasks, Loops, Jobs, Triggers, Bridges, Agents, Marketplace entries, Knowledge, Network channels, and existing session/tab targets.
- Removed the obsolete `searchRoute` and `tasksRoute` helpers.
- Added explicit row ownership and identifiers instead of parsing IDs back out of row keys.
- Added a shared root-owned domain opener used by both root and pushed views. It waits until the runtime workspace and the live window-manager snapshot agree, switches to the owning workspace for global rows, and reports refused/unreachable opens instead of silently falling back.
- Kept Worktree rows on the canonical scope-selection action, including the owning workspace.
- Added Vault `ref` route state so a concrete secret opens in the existing master/detail surface and missing or rejected inventory resolution is shown as an unavailable target.
- Routed installed Extensions through the existing Marketplace detail surface with entry and installed-name identity.
- Added the headline browser scenario and focused unit coverage for identity, similar names, global workspace ownership, stale window-manager state, pushed-view persistence, coordinator refusal, Vault selection, and Vault inventory failure.

## Review

A standalone GPT-5.6-Sol (high reasoning) review was run without the deep-review skill. All five actionable findings were addressed: root-owned pending state for pushed views, truthful QA scope, command/workspace readiness barriers, null coordinator results, and Vault rejected-inventory handling.

## Verification

Passed:

- `bunx turbo run lint typecheck --filter=compozy-web` (zero lint warnings/errors; typecheck and codegen-check passed).
- `bunx turbo run test --filter=compozy-web -- --run src/systems/os/components/__tests__/os-palette-view-stack.test.tsx src/systems/os/hooks/__tests__/use-os-palette-root.test.tsx src/systems/os/lib/__tests__/os-palette-domain-search.test.ts src/systems/vault/hooks/__tests__/use-vault-page.test.tsx src/systems/vault/lib/__tests__/vault-route-search.test.ts` (5 files, 82 tests).
- `git diff --check`.
- Isolated browser walk recorded in `docs/qa/reports/2026-08-24-eng-131-palette-direct.md`, with clean QA teardown evidence. The workspace Loop route and an independent deep link both rendered the same detail page.

`make gate` was run as requested. Its lint, typecheck, codegen, and web execution started cleanly, but the full web lane ended with one unrelated failure in `src/systems/os/hooks/__tests__/use-window-manager-stream.test.tsx` (`Should reconcile a removed workspace before opening the replacement stream`); the other 10 tests in that file passed, and the same failure reproduced in isolation. No unrelated production or test changes were made. `make verify` and `make gate-full` were intentionally not run per the task instruction; CI is expected to provide the full gate.

## Risks and follow-up

- The direct route now depends on the entity identifier and, for global rows, the owning workspace being present in the source projection. Deleted or unavailable targets show an explicit failure instead of a catalog fallback.
- The targeted QA scenario is marked pass only for the walked workspace-Loop journey; the adjacent broad palette scenarios remain explicitly untested rather than being over-claimed.
- The new E2E coverage is in `web/e2e/__tests__/os-shell.spec.ts` and will run in CI.

## Compozy Impact Audit

- **Native tools:** no impact. `compozy__cmd_palette_list`, `compozy__cmd_palette_invoke`, descriptors, schemas, capability gates, and digests are unchanged; only web row projection and routing changed.
- **Extensibility and hooks:** no runtime extension/config lifecycle change. Extension-contributed palette commands retain their existing dispatch path; installed extensions reuse the existing Marketplace detail surface. No `config.toml`, registry, hook, skill, resource, or bridge SDK changes.
- **Workspace data isolation:** in scope and covered. Global rows carry `workspaceId`; the shared opener switches to the owner and waits for both runtime and window-manager workspace identity before opening. Loop routes carry the workspace search parameter, Network/Knowledge keep their workspace parameters, and tests cover foreign-workspace routing and stale command snapshots.
- **Official Compozy skill:** no impact. No public tool ID, CLI path, hook event, capability, memory, or network contract changed, so `skills/compozy/` remains unchanged.

### DIFF
--- a/docs/qa/reports/2026-08-24-eng-131-palette-direct.md
+++ b/docs/qa/reports/2026-08-24-eng-131-palette-direct.md
@@ -1,0 +1,68 @@
+# QA Run Report — 2026-08-24 — ENG-131 command-palette direct entity routes
+
+- **Scope:** Concrete entity rows in the OS command palette now open their detail/action route directly, with workspace ownership preserved.
+- **Cadence tier:** targeted
+- **Build:** `a7a89709d` + ENG-131 working tree · **Environment:** isolated CompozyOS lab, Vite `http://localhost:3031`, daemon `127.0.0.1:60018`
+- **Started:** 2026-08-25T00:47:48Z · **Status:** closed
+
+## Personas
+
+| Persona | Base | Device / Network / Locale | Sessions |
+|---|---|---|---|
+| Sol | isolated workspace `ws_8f9af3086c3bf927` | Chromium / local / pt-BR operator environment | web palette walk |
+
+## Flows in Scope
+
+- `J-command-os-from-palette` — open one concrete workspace loop from the command palette and land on its detail route (`docs/qa/journeys/J-command-os-from-palette.md`).
+- `ET-palette-direct-entity-routes` — verify direct identity, workspace ownership, and deep-link readability (`docs/qa/scenarios/ET-palette-direct-entity-routes.md`).
+
+## Session Matrix & Results
+
+| # | Charter | Journey / Scenario | Persona | Tour | Status | Issue | Fix commit |
+|---|---|---|---|---|---|---|---|
+| 1 | targeted web lab | J-command-os-from-palette / ET-palette-direct-entity-routes | Sol | workspace loop direct route | Pass | — | working tree |
+
+## Session Debriefs
+
+### Targeted web walk — Sol
+
+- **Ran:** 2026-08-25T00:47:48Z → 2026-08-25T00:54:00Z (box respected: yes)
+- **Findings:** The workspace was selected from the shell scope picker. Searching for `palette-direct-loop` displayed the concrete loop row, and selecting it opened `/loops/palette-direct-loop?workspace=ws_8f9af3086c3bf927`. The loop detail rendered its name, contract, graph summary, and run state. A fresh direct navigation to the same route rendered the same detail.
+- **Bugs filed/updated:** none
+- **Scenarios settled:** `ET-palette-direct-entity-routes` → pass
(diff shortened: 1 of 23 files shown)


pedronauck:
CI watch complete (run 32799116684, attempt 2): all frontend, desktop, Go format/lint/build/boundaries, Darwin peer identity, Preflight, PR title, React Doctor (100/100), Vercel, and Greptile checks passed. The two Go race shards still fail on unchanged out-of-scope suites after one retry: shard 2 has 2 failures in internal/daemon TestBootLoopReconciliationBarrier (the daemon never entered the Loop neutralization barrier); shard 1 has 952 internal/store/globaldb failures caused by SQLite migration/checkpoint context deadlines (including checkpoint sqlite wal and migration statements). Verify only propagates GO_TEST_RESULT=failure. No daemon/store files were changed. CodeRabbit was queried in both allowed rounds; it returned no reviews, threads, or actionable issues because the repository hit its included-review rate limit.

PR #516, opened 2026-09-01
fix: preserve profile-scoped extension Agent skills

### LINKED ISSUE #515: Profile-scoped extension Agents leak and lose Agent-local Skill scope
## CompozyOS version

Current `main` at `0bad0f37e219d7bf5a92be86991dda28e68e33be`.

## Operating system

macOS (Apple Silicon).

## Area

- Core / daemon
- CLI
- Agents & providers
- Workspaces
- Extensions / MCP / skills

## ACP agent

Not agent-specific.

## Steps to reproduce

1. Create a resource-only extension that declares an Agent only for a non-default Profile. Give that Agent an Agent-local Skill.
2. Install the extension and enable it for that Profile.
3. Compare `compozy agent list --profile default -o json` with `compozy agent list --profile <profile> -o json`.
4. Register a Workspace and run `compozy skill list`, `skill where`, and `skill view` with `--workspace <workspace-id> --profile <profile> --for-agent <agent>`.
5. Repeat with same-named Agents in the default and non-default Profiles, each with a distinct Agent-local Skill body.

## What happened, and what did you expect?

**Actual:** The Profile-only extension Agent is published into the default Agent catalog. Agent-scoped Skill commands can then fail with `agent not found`, or lose the selected Profile when resolving same-named Agents without a Workspace.

**Expected:** A Profile-scoped extension Agent is visible only through that Profile. Agent-local Skill resolution keeps the selected global, Profile, or Workspace scope and uses the exact Agent definition from that projection, including when another Profile has an Agent with the same name.

## Root cause

The extension Agent/Skill publication path reloads Agent resources from the unprojected manifest after Profile projection. The shared Skill API also resolves by Agent name against filesystem-backed Workspace state instead of passing the concrete projected Agent definition, calls the Workspace catalog for a nil Workspace, and omits the Profile ID from Profile-only scope.

## Suggested fix

- Publish the already-projected static Agent set for each Profile scope.
- Resolve the winning Agent through the catalog for the selected global, Profile, or Workspace scope.
- Pass that concrete Agent definition to the Skill registry.
- Preserve both Profile ID and name in Profile-only resolution.
- Add regressions for default-Profile leakage, global resolution without a Workspace, Profile-only same-name isolation, and Workspace+Profile resolution.

## Logs

```shell
No logs required; the behavior reproduces deterministically through the Agent and Skill CLI surfaces.
```

## Final checks

- [x] I searched open and closed issues and pull requests. Earlier fixes for session command/prompt Skill catalogs do not own Profile projection or the shared `skill list|where|view --for-agent` scope.
- [x] I reproduced the defect on current `main`.
- [x] I am willing to contribute a focused pull request.

richardfogaca:
## What & why

Fixes #515.

Profile-scoped extension Agents now stay inside their selected Profile, and Agent-local Skill lookup uses the exact Agent definition from the active global, Profile, or Workspace projection.

- Publish the already-projected `StaticAgents` set instead of reloading unprojected manifest Agent paths.
- Resolve Agent-scoped Skill queries through the selected catalog lens and pass the concrete Agent definition to the Skill registry.
- Preserve both Profile ID and name for Profile-only Skill scope.
- Keep same-named Agents isolated across Profiles.

An AI coding agent co-wrote this change. Richard Fogaça verified the final diff and validation before publication.

## How you verified it

- `make gate` — pass for the exact branch tree; Go lint and race-enabled affected tests are current.
- `CGO_ENABLED=1 go test -race ./internal/api/core -run '^(TestResolveScopedSkillsUsesProjectedAgentSourcePath|TestProfileOnlySkillScopePreservesProfileIdentity)$' -count=1 -v` — pass for global/no-Workspace, Profile-only, and Workspace+Profile scope.
- `CGO_ENABLED=1 go test -tags=integration ./internal/daemon -run '^TestDaemonE2EImplementTasksShouldCompleteTaskJourney$' -count=1 -v` — pass, including the selected Agent-local Skill path and worker settlement.
- A disposable extension fixture verified that `agent list` does not project a Profile-only Agent into `default`, and `skill list|where|view --for-agent` return distinct Agent-local bodies for same-named Agents in different Profiles.

## Impact

- **CLI and HTTP/UDS:** existing Agent and Skill reads now return the correct scoped resources; no command, flag, route, or payload shape changed.
- **Config and Web:** no `config.toml` key or Web surface changed. No documentation update is needed because the public command contract is unchanged.
- **Native tools:** no tool ID, descriptor, schema, digest, risk flag, or capability gate changed.
- **Extensibility and hooks:** extension Agent publication and Agent-local Skill consumption now use the same Profile projection. No manifest, hook, SDK, MCP sidecar, or lifecycle contract changed.
- **Workspace data isolation:** strengthened by retaining Profile identity through global, Profile-only, and Workspace resolution. No new persisted data was added.
- **Official Compozy skill:** no update required because no public tool ID, CLI path, hook event, capability, or extension resource format changed.

---

- [x] `make gate` passes locally; this PR is delivered only after its required CI checks are green
- [x] New or changed behavior is covered by tests, or I explained above why not
- [x] If an agent wrote or co-wrote this, I named it above and verified the result myself

## Summary by CodeRabbit

- **New Features**
  - Skills can now be resolved directly for specific agent definitions.
  - Agent-specific skills are selected correctly across global, workspace, and profile configurations.

- **Bug Fixes**
  - Profile-based skill scopes now retain the selected profile identity.
  - Extensions now publish configured static agents reliably.
  - Skill resolution better reflects projected agent configurations and prevents incorrect skill assignments.
  - Catalog errors are now reported correctly during skill resolution.


pedronauck (changes requested):
Requesting changes for three missing pieces in the current head:

1. **Add a regression at the daemon publisher's owning layer.** The production change at `internal/daemon/agent_skill_publisher.go:251-269` fixes the original Profile leakage, but the new tests only call `resolveScopedSkills`. Every existing `extensionAgentSkillDeclarationProvider` test omits the profile catalog, so none enters `extensionResourceSnapshotsForProfiles`. Reverting the publisher fix would leave the suite green. Please extend the canonical daemon agent/skill integration suite with a Profile-only extension Agent and prove it is present only in its owning Profile and absent from `default`.

2. **Make the scope matrix actually distinguish default/Profile/Workspace winners.** In `skills_scope_test.go`, `defaultAgent` is stored under `profile-default`, but `ListAgents` always returns `globalEntry`, so that Agent is never selected. The Workspace+Profile case also indexes only `ProfileID` and returns the exact same `work` entry as the Profile-only case. A resolver that ignores default-Profile or Workspace precedence would still pass. Use distinct source paths/bodies for global, default Profile, non-default Profile, and Workspace+Profile, and assert each public scope selects its own winner.

3. **Close the QA tracker loop for this user-visible behavior change.** This PR changes the observable results of `agent list` and `skill list|where|view --for-agent` across Profiles, but it neither adds/resets an owning `docs/qa/scenarios/` entry nor records a real isolated walk. Add a content-addressed scenario (or reset the exact owning scenarios), exercise the default/Profile/Workspace same-name isolation through the public CLI/API surfaces, and record the verdict and teardown evidence.

The focused core tests pass, but they do not cover the original publisher regression or the claimed scope-precedence matrix above.

pedronauck (approved):

PR #662, opened 2026-09-20
fix: advance role fallback when the provider refuses an accepted turn

### LINKED ISSUE #617: Let a fallback route name an account, so a session refused for quota can advance to the operator's other seat
## What problem would this solve?

**Context, since this arrives without one.** I ran a structured comparison of agent-orchestration tools
across the same set of axes, and CompozyOS was one of them. That is why the numbers below are exact and
why several reports from me land close together. They came from one measurement pass on a dedicated
Linux box, not a production incident. Happy to re-run any of it differently if that helps.

**A fallback route can change provider and model, but not the account. Quota is exhausted per account.**

### What CompozyOS already does well here, measured

Both of my provider accounts hit their session limit in the same window, during a real run. This was not
simulated with a broken credential; it was the actual limit.

```
account A, 00:22 -> {"code":-32603,"message":"Internal error: You've hit your session limit
                     · resets 3:30am (America/Sao_Paulo)","data":{"errorKind":"rate_limit"}}
                    provider_failure_kind=rate_limited
                    next_action=retry
                    guidance=retry after the provider recovers
account B, 00:26 -> the same message, the same reset time
```

**This is the most precise failure reporting I measured anywhere.** It names the failure kind,
prescribes an action, and gives the reset time. Of the four tools that behaved comparably, one reported
"rate limited" for what was actually a 401, one said nothing about the cause, and CompozyOS named it.

It also stopped spending immediately, which is the right thing.

### The gap

**The operator has a second account configured, and the runtime has no declared way to use it.**

I switched accounts by hand: `compozy config set providers.claude.command ...`, restart, rerun. That
worked. It was my hands.

The documentation states this as current behavior, so I am not reporting a surprise:

> *"Authentication and rate-limit failures end the affected turn while keeping the session available.
> After restoring authentication or waiting for the provider to recover, send another prompt in the same
> session. CompozyOS does not automatically retry the failed prompt."*
> Source: `docs/agents/providers.mdx`

### Why I think the pieces are already there

`fallback_chain` already exists, is documented, is observable, and already has the safety boundary this
would need:

```toml
[[roles.auto_title.fallback_chain]]
# Fallback routes are tried in declaration order before a session is accepted.
provider = "cursor"
model = "grok-4.6"
reasoning_effort = "high"
```

> *"A session-backed consumer may advance only before ACP accepts a session; CompozyOS never silently
> reroutes work after acceptance. Immediately before each fallback attempt, CompozyOS emits
> `role.fallback.used` with the role."*
> Source: `docs/configuration/config-toml.mdx`

**That boundary is exactly right, and it is why I think this is a small extension rather than a new
subsystem.** Two things block it from covering quota:

**1 · A fallback route cannot name an account.**

```go
// internal/config/roles.go:105-111
type RoleFallback struct {
    Provider        string
    Model           string
    ReasoningEffort string
    Speed           speedpkg.Speed
    ACPOptions      []ACPOptionSelection
}
```

There is no `command`. The account is selected by `command`, either globally through
`providers.claude.command` or per agent through `agent --command`. So a chain can send work to a
different provider or a different model, but not to the operator's other account on the same provider.

**2 · Only background roles have a chain.**

`FallbackChain` appears on `Coordinator`, `Dream`, `CheckpointSummary`, `MemoryExtractor`, `AutoTitle`
and `MemoryController` (`internal/config/roles.go:69-74`). It does not appear on `AgentDef`
(`internal/config/agent.go`), whose routing fields are `Provider`, `Command`, `Model`,
`ReasoningEffort` and `Speed`. So the sessions that do the actual work cannot fall back at all.

### And the action vocabulary has no word for it

When a provider fails, CompozyOS prescribes one of nine actions
(`internal/acp/provider_failure.go:33-41`):

```
install_cli   login   bind_secret   inspect   change_model
request_permission   wait   retry   no_retry
```

`bind_secret` means "go configure a credential" and `inspect` means "go look". Neither means "use the
other account you already configured". `rotate_credential`, or any equivalent, does not appear anywhere
in the tree on `0.3.0-beta.25`.

---

## What would you like to happen?

**Let a fallback route carry an account, and let work sessions reach a chain.**

Two independent pieces, and the first is useful on its own:

**1 · Give `RoleFallback` a way to name an account.** Either a `command` field, matching how accounts are
already selected everywhere else, or a named credential reference if you prefer not to put a command
string in a fallback table. This alone would let background roles survive one account's limit.

**2 · Let `AgentDef` declare a chain too**, so a user session that is refused before ACP acceptance can
advance to the next route.

**Keep the boundary you already declared.** Advance only before ACP accepts a session, never reroute
after acceptance, and keep emitting `role.fallback.used` (or an agent equivalent) before each attempt so
the operator can see which account did the work. A silent account switch would be worse than the current
behavior, because billing and rate limits are per account and the operator needs to know.

**Optionally, a tenth action.** If the chain can now resolve, then `rate_limited` could prescribe
something like `use_fallback` instead of `retry` when a route is available. That is cosmetic next to the
two pieces above.

---

## Area

`Agents & providers`, `Core / daemon`, `Sessions`

---

## Alternatives you considered

| alternative | why it does not close the gap |
|---|---|
| **the operator switches by hand** | that is what I did: `config set` plus restart plus rerun. It works, and it is the thing an orchestrator exists to remove |
| **`change_model` to a cheaper model** | already in the vocabulary, and it does not help: the limit is on the account, not the model |
| **a second provider in the chain** | already possible today, and it changes vendor. I wanted the same vendor, my other seat |
| **wait for the reset** | the prescribed action today. Correct, and it costs the rest of the window |
| **run each agent pinned to a different account** | possible with `agent --command`, and it partitions capacity in advance instead of absorbing a limit when it happens |

**Prior art in the tree:** `fallback_chain` for roles, with ordered routes, a declared acceptance
boundary, and a `role.fallback.used` event. This request reuses all three.

**Nearest existing issues, and why this is not a duplicate:**

| issue | why it is different |
|---|---|
| **#564**, closed, mine | spawned sessions ignored the creating agent's `command`. Same axis, about propagating an account that was already chosen. This one is about choosing a different account when the first is refused |
| **#570**, closed | Loop session policy dropping the owning Profile |
| **#229** and **#473**, closed | model routing by price and by complexity, not account capacity |

**Duplicate search performed** (title and body, open and closed, `is:issue`): `fallback_chain` ·
`rate limit fallback` · `rotate credential` · `second account` · `quota exhausted` ·
`provider account switch` · `session limit` · `multiple accounts` · `provider failure action`.

---

## Contribution

- [ ] I'm willing to open a pull request for this.

jillesmc:
Refs #617. **This PR does not close that issue** — it removes a prerequisite. See "What this does not do".

## The gap

A role fallback chain was only consulted when the session object failed to materialize:

```go
// internal/daemon/role_fallback.go
value, accepted, err := invoke(ctx, primary)
if accepted { return value, err }   // accepted == "a session exists"
```

A provider that **accepts the session and then refuses the turn** never reaches the chain. That is the
shape both refusals in the issue arrive in — a quota refusal (`-32603`, `errorKind: rate_limit`) and the
credential refusal the reporter actually measured (`-32000 Authentication required`). The session is
created, the chain returns, and the refusal only surfaces during the prompt.

```mermaid
flowchart LR
  subgraph before["before — accepted = spawned != nil"]
    A1[Spawn] -->|session exists| A2[accepted = true]
    A2 --> A3[chain returns]
    A3 --> A4[Prompt]
    A4 -->|provider refuses| A5[operation fails<br/>second route never runs]
  end
  subgraph after["after — accepted = the provider took the turn"]
    B1[Spawn] --> B2[Prompt + collect]
    B2 -->|refused, no output| B3[stop child<br/>accepted = false]
    B3 --> B4[next route runs]
    B2 -->|any other outcome| B5[accepted = true<br/>unchanged behaviour]
  end
```

## Strategy: align with the role that already worked

`BreakTie` (memory controller) **already advanced on a refused turn**, because `InvokeTransientModel`
sets `Accepted` only after the prompt succeeds (`internal/session/transient_model.go:118-121`). It was
the only one of five callers doing so. This PR makes `Generate` (auto-title) and `Summarize`
(checkpoint summary) match it — they now prompt inside the invocation closure instead of after it.

Classification needs no new vocabulary: `AgentEvent` already carries a typed
`ProviderError.Code` ∈ `{provider_auth_required, provider_rate_limited}`, built for exactly these two
cases in `internal/acp/provider_error_event.go:27-47`. The new predicate is nine lines and reads it.

| role | closure covers | acceptance predicate | changed |
|---|---|---|---|
| `Generate` (auto-title) | spawn + prompt | provider took the turn | **yes** |
| `Summarize` (checkpoint) | create + prompt | provider took the turn | **yes** |
| `BreakTie` (memory controller) | full call | `result.Accepted` | no — already correct |
| `startCoordinatorSession` | create only | `spawned != nil` | no — hands the session on |
| `spawnExtractorSession` | spawn only | `spawned != nil` | no — hands the session on |

**The declared boundary is preserved.** `docs/configuration/config-toml.mdx` says CompozyOS never
silently reroutes work after acceptance. A route advances only when the turn produced **no output** —
that is retrying work that never started, not rerouting work in flight. Every other prompt error stays
on the accepted route and fails exactly as before. `role.fallback.used` still precedes every attempt.

## Findings worth reviewing

Three claims in the issue did not survive verification against the graph, and they changed the design:

1. **Five callers, not three.** `invokeRoleWithFallback` has five production callers. The issue names three.
2. **The predicate was never uniform.** `BreakTie` already had the wider trigger. The fix is alignment, not invention.
3. **`fallback_chain` on `AgentDef` would be dead config.** User prompts — the ones that exhaust quota —
   never reach `role_fallback.go`. They go through the session prompt-failure barrier, which classifies
   on `store.FailureKind ∈ {FailureProcess, FailureTransport}` (`manager_prompt_recovery.go:37-47`,
   `manager_prompt.go:79-89`). A rate limit is neither, so the turn ends and the session stays alive.
   Adding a chain to `AgentDef` would attach it to a struct no user-prompt path consults.

A **regression found during the impact audit and fixed in this PR**: the first draft collapsed the spawn
result to `if spawned == nil`, which silently dropped a non-nil error returned alongside a live session.
That is a reachable path — `Spawn` returns `(child, err)` when spawn-created hook dispatch fails
(`internal/session/spawn.go:131`). Swallowing it would have let a denied hook through. Both roles now
fail the attempt on any spawn error, as the original code did.

## What this does not do — and what closes #617

**User sessions still have no quota fallback**, which is the issue's actual motivation. The two roles
fixed here are background work — session titles and compaction summaries. Nobody's work is blocked by
them and they are not what exhausts an account. This PR makes the fallback mechanism consistent and
correct; it does not give the operator's own session a second seat.

The follow-up is designed, and it reuses machinery that already exists rather than adding a subsystem:

1. `fallback_chain` on `AgentDef`, so a user session has routes at all.
2. `promptEventIsRecoverable` admits a typed provider refusal with no execution evidence — the same
   condition this PR enforces for roles. `internal/session` currently never reads
   `AgentEvent.ProviderError`, though the typed code is already on the event.
3. `recoverPromptRuntime` (`internal/session/manager_prompt_recovery.go:241`) already relaunches the
   runtime under a live session and replays the turn. It passes `snapshot.selection` today and would
   pass the next route's selection instead; `preparePromptRuntimePlan` already accepts an arbitrary
   `RuntimeSelection`.
4. Consent model: a route declared in config **is** the consent — the runtime advances and emits the
   event, so an overnight refusal does not park the work. With no declared route, a durable pending
   interaction asks the operator, reusing `session_pending_interactions`, which is already durable,
   resumable and exposed on every surface. An accept is scoped to the provider-reported reset time;
   a timeout fails the turn and never advances on its own.

That half changes a public config surface and a persisted enum, so it belongs in its own change with
its own compatibility and QA treatment.

Also deliberately not included, per the reporter's own correction on the issue:
`command` on `RoleFallback` (a provider overlay already names an account), `fallback_chain` on
`AgentDef` (see finding 3), and a tenth `use_fallback` action.

## Review guide

- **Start here:** `internal/daemon/role_fallback.go` — `providerRefusedTurnError`, the whole new predicate.
- **Then:** the two closures in `auto_title_generator.go` and `checkpoint_summary_summarizer.go`. The
  question to ask each is *"which outcomes return `accepted = false`?"* — the answer must be only a
  provider refusal, never a generic error.
- **Lifecycle:** `child` / `summarySession` is assigned only on the accepted path, so the outer stop
  defer fires for exactly one session; a refused attempt stops its own child before returning. An
  exhausted chain leaves nothing running.
- **Not changed:** no route, verb, tool ID, DTO, config key, hook, SQLite shape or migration. No public
  signature changed — the generic's type parameter moved from `*session.Session` to `string`, which is
  internal to each call.

## Verification

- `make gate`: `go-lint` 0 issues, `go-test ./internal/daemon/...` green under `-race` (136s).
- Tests in `role_fallback_test.go`: a predicate table (rate-limited and auth-required mark; an
  unclassified agent error does not) plus one end-to-end advance through the real
  `forkedAutoTitleGenerator`, reusing the existing `autoTitleSpawnSessionsStub`.
- Both were **mutation-checked**: with the predicate stubbed to never mark a refusal, the table fails on
  the rate-limited case and the end-to-end test fails with the primary route's error, proving neither is
  vacuous.
- Impact audit over the changed symbols: `providerRefusedTurnError` 9 impacted / LOW, reaching only
  `autoTitleRuntime.process` and `CheckpointSummaryService.summarizeCheckpoint`. No taint findings in
  the touched files — with the caveat that closure flows are not modelled and this logic lives in a
  closure, so that is not proof.

Cross-surface impact is recorded in `docs/_memory/change-impact.md`.

## Summary by CodeRabbit

* **New Features**
  * Automatic titles and checkpoint summaries can continue through configured fallback providers when a provider refuses before producing output.
  * Refused provider sessions are stopped before the next fallback is attempted.
  * Generated Graphify and GitNexus output is excluded from version control.

* **Bug Fixes**
  * Improved handling of provider refusals, session cleanup failures, session creation failures, partial output, and empty generated output.


jillesmc:
## Review remediation (749381363)

Both reviewers independently flagged the same two P1s. Both were valid, and both broke the boundary this PR's description claims to preserve — thanks, that was the useful kind of catch.

**1 · Refusal after output began** (greptile `auto_title_generator.go:261`, coderabbit same site)

The predicate keyed on the error classification alone, ignoring whether the agent had already streamed text. A provider that produced output and then refused advanced the chain — precisely the *"never silently reroutes work after acceptance"* case. Both collectors now mark a refusal only while the accumulated output is empty; once the turn produced anything, the error stays on its accepted route.

**2 · Advancing past a failed cleanup** (greptile `auto_title_generator.go:128`, coderabbit `#L127-L129`)

`errors.Join` preserved the refusal marker, so a failed stop still advanced and could start a route beside a session that was never stopped. The attempt now keeps ownership of that session (assigned to `child` / `summarySession`) so the outer stop runs, and returns accepted-with-error instead of advancing.

**3 · Test convention** — table entries and the end-to-end scenario now run as `t.Run("Should …")` subtests.

New regression test for #1: *"Should stay on the accepted route when the turn already produced output"*. Mutation-checked — it fails when the output guard is removed. `Generate` crossed `funlen` after the fix, so the spawn options moved to `autoTitleSpawnOpts`.

**Declined:** the nitpick asking for an equivalent end-to-end refusal test on the checkpoint-summary path. Test scope on this PR was calibrated deliberately: the checkpoint path is the same predicate and the same wiring as auto-title, so a second stub buys duplicated maintenance rather than independent coverage. If the two paths ever diverge, that judgement should be revisited.

`make gate` green on the new head: `go-lint` 0 issues, `go-test ./internal/daemon/...` under `-race` (139s). `docs/_memory/change-impact.md` records the tightened boundary.

jillesmc:
@coderabbitai review

The three actionable findings were addressed in 749381363 (details in the previous comment). The standing `CHANGES_REQUESTED` is pinned to 4d5da7a03 and predates the fix — please re-review the current head.

